PR #76, opened 2015-03-02
(tile_map) Fixing conflict between Boost signals and QT.

elliotjo:
Looks like this was fixed in the hydro branch but not master.


elliotjo:
Added another commit to improve the marker_plugin support for pose/orientation.

PR #116, opened 2015-06-30
GPSFix plugin

pjreed:
Adding a plugin to display GPSFix messages.


mapviz_plugins/src/gps_plugin.cpp
```diff
@@ -0,0 +1,486 @@
...
+    StampedPoint stamped_point;
+    stamped_point.stamp = gps->header.stamp;
+
+    double x;
+    double y;
+    local_xy_util_.ToLocalXy(gps->latitude, gps->longitude, x, y);
+
+    stamped_point.point = tf::Point(x, y, gps->altitude);
+
+    stamped_point.orientation = tf::Quaternion(
```
malban: The GPS track value can be used here to get the orientation so the arrow view will display correctly.

PR #130, opened 2015-07-01
Renaming "mutlires" to "multires".

pjreed:
The definition of the multires image plugin accidentally called it "mutlires_image" instead of "multires_image".


malban:
Can you add a hack to make config files with the old wrong name still load the plug-in.

PR #248, opened 2015-10-09
Feature/gps plugin jade

cbandera:
I adapted the gps plugin to the new message format in jade.

Therefor I have reverted to earlier commits, where the plugin file was removed. 

NOTE: The orientation is not longer a part of the GPS msg (NavSatFix.msg) therefor I had to constantly set it to 0.


evenator:
I'm not sure if we should take this approach or split NavSatFix out into its own plugin (with both plugins available in indigo and GpsFix removed in jade).

evenator:
This looks good, but please remove the "arrows" draw style, as the arrows have no meaning without heading. Once that's done, we'll test this and merge it in.

Thanks!

cbandera:
@evenator Hey, I removed the arrows and renamed the plugin. See if you like it.

evenator:
@malban

PR #253, opened 2015-10-16
A plugin for publishing clicked points.

pjreed:
This actually adds two useful things.  The first is a new method, MapCanvas::MapGlCoordToFixedFrame, that will convert a pixel coordinate to its equivalent position in mapviz's current fixed frame.

It also adds a plugin that listens for clicks on the GL canvas, then publishes the locations of those clicks in a specified frame to a specified topic.


cbandera:
Hi, thank you for your plugin. I had a hard time finding it, as there is no documentation on it yet.
It would be awesome, if you could add that information to the mapviz plugin list in the readme and the ROS wiki.
Also it would be great, if one could publish the point in gps coordinates.

pjreed:
I'll add documentation for it to the README.  I don't have an account on the ROS wiki and it looks like registration is disabled at the moment, so I'll see if I can get somebody else to do it.

For publishing the point in GPS coordinates, the easiest way is, if you have tf frames being published in that coordinate frame ("/wgs84", for example), set that as the "Frame:" option for the plugin.

PR #261, opened 2015-10-29
Add OpenStreetMap as source for tile_map

vincentrou:


vincentrou:
You are right, it is not allowed.
I was not aware of this [policy.] (http://wiki.openstreetmap.org/wiki/Tile.openstreetmap.org/Usage_policy)
"OpenStreetMap data is free for everyone to use. Our tile servers are not."

Issue #283, opened 2016-01-07
Travis configuration not in jade-devel branch


evenator:
jade-devel builds are failing on travis because the .travis.yml file isn't present in that branch.

PR #285, opened 2016-01-07
Adding Jade support

pjreed:
In addition to merging in the status indicator buttons from the indigo-devel branch, this also modifies the Travis CI script to build this branch using Jade.


pjreed:
This should fix #283 .

PR #286, opened 2016-01-07
Adds SelecTopicDialog to mapviz.

elliotjo:
This commit adds the SelectTopicDialog that can be used in plugins to
provide the user with a dialog to select topics.  Typically we have
done this with a lot of duplicated code across all the plugins.  This
commit also updates the plugins in mapviz_plugins to use the new
dialog.

The new dialog provides several benefits:

- Reduced code duplication
- Simplifies writing new plugins
- Common behavior between all plugins
- Topics sorted by name
- User can filter topics by substring
- Continuously checks the master for new topics while the dialog is open.


elliotjo:
Arg! It is SelectTopicDialog, not SelecTopicDialog.  I made the same mistake while writing the code too.

PR #287, opened 2016-01-08
Adding color button widget and updating plugins.

elliotjo:
This commit adds a subclass of QPushButton called ColorButton that
implements a widget for displaying and selecting colors.  We've been
doing this manually everywhere with duplicated code.  This is a simple
abstraction but allows us to elminate a lot of duplication, especially
in plugins that have multiple color selections.


elliotjo:
This merge should not change anything from a user's point of view, including currently saved settings.

PR #293, opened 2016-01-14
Initialize mapviz as an anonymous node.

elliotjo:


pjreed:
This works when you run mapviz through rosrun, but when you launch mapviz through roslaunch ("roslaunch mapviz mapviz.launch"), the node is still just named "mapviz".  Is it worth also modifying that?

elliotjo:
I've added a change to fix the launch file too.

PR #294, opened 2016-01-15
add pointcloud2 plugin

nick-alton:
Adds point cloud 2 plugin with dynamic field interface, and automatic min and max settings.


pjreed:
I'm having a few issues getting this to work.  I'm building this with the latest versions of marti_common and marti_messages and testing it by playing the cone_course bag file (ask Jason for it if I'm not around).  I can draw odometry, GPS, and map tiles in mapviz without any issue, and I can see the points on /lidar/vh_wm_points and /lidar/vh_wm_points_ground in rviz.

However, when I add a pointcloud2 plugin in mapviz, I don't see any point from the point cloud at all in my display.  Its status says "OK", so I assume it's getting messages.  Maybe it's not transforming them into the proper frame?  My mapviz fixed frame is near_field and target frame is veh_near_field.

Also, there's some odd behavior with the Use Rainbow / Auto Min Max checkboxes.  Checking Auto Min Max hides the color selection boxes, but it should probably still be possible to change those even when the limits are being automatically determined.  It would probably also be better to simply disable the related fields than remove them; removing them causes the size of the widget to change, which causes some layout issues when the plugin panel is minimized, re-expanded, and then the fields are shown again.

There may also be some issues with saving and loading the widget states.  If I check both of the checkboxes and close mapviz, when I re-open it, the Rainbow checkbox is not visible, and the minimum color box is visible, but none of the other widgets.  Unchecking the Auto Max/Min box causes the widgets below it to reappear, but nothing seems to make the Rainbow checkbox reappear.

nick-alton:
Alright. Thanks for looking at it I'll tinker with it some more on Monday and see if I can't fix it. I did have some issues with mapviz where it would load and constantly spam "OK" into the terminal usually I just played with the selected frames and it would fix itself when I went back to the near_field. I will look into disabling the input fields as opposed to vanishing them, I did notice some of that behavior with the buttons at times too.

pjreed:
Looking better after the most recent push, but there's still a little odd behavior.  It appears that unchecking "Use Auto Max/Min" shows all of the setting widgets, including the color boxes.

So, if you do this:
1. Start with "Use Rainbow" and "Use Auto Max/Min" unchecked; all widgets should be visible.
2. Check "Use Auto Max/Min".
3. Check "Use Rainbow".
4. Uncheck "Use Auto Max/Min".
5. Now, all of the widgets, including the color boxes, are visible.

mapviz_plugins/mapviz_plugins.xml
```diff
@@ -38,6 +38,9 @@
   <class name="mapviz_plugins/laserscan" type="mapviz_plugins::LaserScanPlugin" base_class_type="mapviz::MapvizPlugin">
     <description>LaserScan mapviz plugin.</description>
   </class>
+  <class name="mapviz_plugins/pointcloud2" type="mapviz_plugins::PointCloud2Plugin" base_class_type="mapviz::MapvizPlugin">
+    <description>LaserScan mapviz plugin.</description>
```
pjreed: Update the description for the plugin here.

mapviz_plugins/include/mapviz_plugins/pointcloud2_plugin.h
```diff
@@ -0,0 +1,155 @@
...
+      {
+        COLOR_FLAT = 0,
+        COLOR_X = 1,
+        COLOR_Y = 2,
+        COLOR_Z = 3,
+        EXTRA_1 = 4,
+        EXTRA_2 = 5,
+        EXTRA_3 = 6,
+        EXTRA_4 = 7
+      };
```
pjreed: It looks like here and in a number of other places, the plugin is limited to handling a maximum of 4 different feature types (other than x/y/z); would it be possible to handle an arbitrary number of features?

evenator: +1 for arbitrary number of feature types

Issue #268, opened 2015-11-22
The canvas update should happen at a fixed rate and not need to be triggered by plug-ins.


malban:
This will remove the need for plug-ins to trigger updates and avoid related bugs.

It may also make the video capture rate more stable.

PR #299, opened 2016-01-20
Update map canvas at a fixed rate.

elliotjo:
Addresses Issue #268.

This update adds a timer to the map canvas to repaint at a fixed rate.
The default rate is 50 Hz, but there is a method to change it (not
exposed to the UI at the moment).  50Hz was chosen because it is fast
enough to give smooth animations and we almost always are running
mapviz with at least one plugin triggering updates from a 50Hz topic.

This should be a safe merge because the update() method is still exposed, so existing plugins won't break.  But I think it would be good to have a few other people use it for a day or two before it gets merged in just in case.

### DIFF
--- a/mapviz/include/mapviz/map_canvas.h
+++ b/mapviz/include/mapviz/map_canvas.h
@@ -46,6 +46,7 @@
 #include <QMouseEvent>
 #include <QWheelEvent>
 #include <QColor>
+#include <QTimer>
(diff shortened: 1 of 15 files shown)


pjreed:
Looks like everything is working for me.  I've also tested it with some other plugins that still call update() and they're fine, too.

PR #303, opened 2016-01-27
QPainter support for plugins

pjreed:
I started work on this a while back, but I finally got around to polishing it up a bit.  The order of plugins is now properly respected; for each individual plugin, GL drawing will still be done before QPainter operations, but QPainter operations will still be done before the next plugin's GL drawing, so the layer ordering looks right.

In a nutshell, this adds a new method that MapvizPlugins can implement, MapvizPlugin::Paint, which allows them to use a Qt QPainter object to draw on Mapviz's canvas with the same coordinate system that they would normally do GL drawing.  This is handy because the QPainter provides a number of useful methods for drawing text, textures, and a variety of shapes using complex pens and brushes that are all a pain to do with native OpenGL commands.  The Paint function has an empty default implementation, so the API has not changed and existing plugins will work without modification.

### DIFF
--- a/mapviz/include/mapviz/map_canvas.h
+++ b/mapviz/include/mapviz/map_canvas.h
@@ -139,7 +139,7 @@ namespace mapviz
   protected:
     void initializeGL();
     void resizeGL(int w, int h);
-    void paintGL();
+    void paintEvent(QPaintEvent* event);
     void wheelEvent(QWheelEvent* e);
     void mousePressEvent(QMouseEvent* e);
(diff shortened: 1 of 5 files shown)


pjreed:
I modified the Marker plugin so that it will use a QPainter to draw text markers.  This doesn't really add any functionality for the user; it's just intended as an example of how to use the QPainter in a plugin.

It would also be really easy to do something like add a black outline around the text to make it stand out more, but the paint surface is not currently anti-aliased, and aliased black borders around text look really awful.

PR #304, opened 2016-01-27
Adding the ability to toggle anti-aliasing

pjreed:
Now there's a checkbox under the "View" menu that will toggle whether
anti-aliasing is applied to the canvas.  In some situations this will
make the display look much prettier at a slight performance cost.


pjreed:
Before and after screenshots:
mapviz_20160127t231559_675467
mapviz_20160127t231604_524579

PR #305, opened 2016-01-28
Enable QPainter antialiasing

pjreed:
This depends on #303 and #304 and should be merged after both of them.

After turning on QPainter support, there are a few extra flags that have to be set for anti-aliasing to work with it.


pjreed:
Going to remake this merge request to work around conflicts.

PR #312, opened 2016-02-02
Removing extraneous calls to MapCanvas::update()

pjreed:
Now that update() is being called automatically at a rate of 50 Hz, the explicit calls in many locations are unnecessary.  It was also possible for it to be called in some of these locations from a non-main thread, which is invalid and could cause crashes.

Also, there was an extra call to glPopMatrix() that did not have a matching glPushMatrix().  It didn't seem to be hurting anything, but I removed it.


evenator:
We should also port this to Jade

Issue #317, opened 2016-02-04
Plugins after the first one sometimes don't draw


pjreed:
PR #306 introduced a bug that caused the OpenGL viewport to become un-transformed properly after the first plugin was drawn, which could make subsequent plugins not render properly.

PR #318, opened 2016-02-04
Fixing #317; weird plugin transformation glitches

pjreed:
First, the model view matrix needs to be saved and restored around QPainter operations because Qt clears several GL variables.  Also, the image plugin needed to explicitly call glMatrixMode(GL_PROJECTION); it does a few operations on the projection matrix and was just assuming that was the current matrix mode.  Also, I added a function that plugins need to override if they want to do QPainter operations; this will eliminate unnecessary overhead for plugins that do not.

### DIFF
--- a/mapviz/include/mapviz/mapviz_plugin.h
+++ b/mapviz/include/mapviz/mapviz_plugin.h
@@ -230,6 +230,15 @@ namespace mapviz
 
     void SetIcon(IconWidget* icon) { icon_ = icon; }
 
+    /**
+     * Override this to return "true" if you want QPainter support for your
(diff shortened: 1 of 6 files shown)


elliotjo:
I wonder if it would be a good idea to push the current transform matrix
and all opengl attributes before calling a plugin's draw method and then
restore them afterwards to help isolate the plugins better.

pjreed:
That's a good idea.  I've modified the code to save and restore all matrices and attributes around each plugin.  A plugin could still mess things up by popping more or less often than they push, but this should still help to keep them isolated from each other.

PR #328, opened 2016-02-15
Image issue

nick-alton:
Added functionality that unsubscribes/subscribes an image or disparity topic when the respective plugin is hidden.


pjreed:
It looks like this is properly unsubscribing when an image topic is hidden, but if the image plugin is initialized hidden, it subscribes anyway.

PR #329, opened 2016-02-16
fixing a variable name

nick-alton:
Changed a variable named 'hidden' inside  of a function within the disparity and image plugins to 'visible' to prevent any confusion.


evenator:
LGTM. I'll merge it when the CI build passes.

PR #331, opened 2016-03-03
Route plugin

nick-alton:
Route Plugin release version. Has route drawing along with a RoutePosition marker subscriber.


evenator:
I just cut releases of `marti_messages` and `marti_common`. I expect the CI on this will pass as soon as those get built by the OSRF build farm.
In the mean time, I'm going to go ahead and merge this, because the build works for me and @jatowler when built against `indigo-devel` of `marti_messages` and that's good enough for me.

Issue #325, opened 2016-02-11
OpenGL blending is enabled but overwritten.


elliotjo:
OpenGL blending (alpha channel support) is enabled by default (in MapCanvas::initializeGL) but was broken by the support for QPainter.  It seems like there is some bad interaction with the call to QPainter::beignNativePainting() in MapCanvas::paintEvent().  Placing a call to glEnable(GL_BLEND) after the line restores functionality in plugins, but before does not.  This is just a work-around however, we should find the actual problem in case other GL state variables are being overwritten.

pjreed:
Are there any existing plugins in which this behavior can easily be demonstrated?

pjreed:
I believe that alpha transparency not working in the costmap plugin is a side effect of this.
@jgassaway @nick-alton

PR #332, opened 2016-03-24
Fixing blending for GL drawing

pjreed:
Fix for #325. The call to QGLWidget::beginNativePainting has a side effect of clearing GL settings related to blending and depth testing, and that was causing alpha transparency to not work right for plugins.  I fixed it by manually re-enabling those settings every time beginNativePainting is called.


pjreed:
Note that mapviz_plugins fails to build because the version of marti_messages in the public repositories does not yet have the Route messages in it.   We'll need to deploy a new version of marti_messages to fix it.

Issue #265, opened 2015-11-14
Changing frames in subsequent marker messages not handled correctly.


malban:
Changing frames in subsequent marker messages not handled correctly.

The data appears to be displayed as if it was with respect to the frame of the first message.

pjreed:
I looked into this, and the cause is pretty obvious; the plugin sets its source_frame_ member variable the first time a message is received, and mapviz uses that to transform the contents, and its value is never changed again unless you change the topic name.

It looks like there are actually several plugins that have this same behavior, so I'll fix all of them at once.

PR #334, opened 2016-04-07
Fix for #265; message source frames don't update (Indigo)

pjreed:
Several plugins were storing the source frames of messages received when
they first received a message but never updating them, so subsequent
messages in different frames would be rendered incorrectly.

### DIFF
--- a/mapviz_plugins/src/route_plugin.cpp
+++ b/mapviz_plugins/src/route_plugin.cpp
@@ -243,11 +243,11 @@ namespace mapviz_plugins
     {
       source_frame_ = "/wgs84";
     }
-    unsigned long route_size = route->route_points.size();
+    size_t route_size = route->route_points.size();
 
     points_.clear();
 
-    for (unsigned long i = 0; i < route_size; i++)
+    for (size_t i = 0; i < route_size; i++)
     {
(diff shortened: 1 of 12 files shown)


evenator:
Per our discussion at lunch on Friday, this needs a more sophisticated fix for any plugin that keeps a buffer of messages.

pjreed:
I've updated all of the plugins that store buffers of messages so that they store the source frames for each message individually and transform them appropriately, so that should fix the issue.

mapviz_plugins/src/route_plugin.cpp
```diff
@@ -232,22 +232,22 @@ namespace mapviz_plugins
...
-    int route_size = route->route_points.size();
+    if (!route->header.frame_id.empty())
+    {
+      source_frame_ = route->header.frame_id;
+    }
+    else
+    {
+      source_frame_ = "/wgs84";
+    }
+    unsigned long route_size = route->route_points.size();
```
evenator: I'd prefer size_t over unsigned long (and below on line 250 as well).

evenator:
I made one comment inline about size datatypes. Otherwise, LGTM. I vote to merge this, then soak-test for a while before we cut another Debian release.

PR #335, opened 2016-04-07
Fix for #265; message source frames don't update (Jade)

pjreed:
ROS Jade version of #334.


evenator:
Per our discussion at lunch on Friday, this needs a more sophisticated fix for any plugin that keeps a buffer of messages.

Issue #339, opened 2016-04-07
Depend explicitly on OpenCV 2


pjreed:
Mapviz, mapviz_plugins, and multires_image were all developed against OpenCV 2, but their CMakeLists only specificy "OpenCV REQUIRED", so they will attempt to compile against OpenCV 3 if it's installed.  They should be modified to specify the version number.

PR #341, opened 2016-04-07
Fix for #339; explicitly depending on OpenCV 2 (Jade)

pjreed:


evenator:
LGTM and compiles against shadow-fixed.

PR #343, opened 2016-04-24
Implement service for adding and modifying mapviz displays.

malban:
Implements service for adding and modifying mapviz displays.

The service gives control over:
 - display type
 - name
 - draw order
 - visibility
 - plug-in specific parameters

If an existing display with the same type and name exists, it will be updated based on the parameters in the service, otherwise a new display is added.


mapviz/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(mapviz)
 
+set(CMAKE_CXX_FLAGS "-std=c++11 ${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS")
```
elliotjo: Consider making these flag changes local to the mapviz target instead of global to all packages in the workspace:

set_target_properties(mapviz PROPERTIES COMPILE_FLAGS "-std=c++11 -D__STDC_FORMAT_MACROS")

mapviz_plugins/include/mapviz_plugins/attitude_indicator_plugin.h
```diff
@@ -93,6 +93,7 @@ class AttitudeIndicatorPlugin : public mapviz::MapvizPlugin
 
   void Transform() {}
 
+  virtual void UpdateConfig(std::map<std::string, std::string>& params);
```
elliotjo: virtual keyword is not needed here or the other plugins since they aren't intended to be sub-classed.  It doesn't hurt anything, but it is inconsistent with the other overload declarations.

mapviz/include/mapviz/mapviz_plugin.h
```diff
@@ -220,6 +220,7 @@ namespace mapviz
 
     virtual void Transform() = 0;
 
+    virtual void UpdateConfig(std::map<std::string, std::string>& params) {}
```
elliotjo: Could we modify to swri_yaml_util to add a function to parse a YAML string?  You could easily pack the key value pairs in an AddMapvizDisplay request into a simple YAML dictionary, parse it, and then pass it to the normal LoadConfig instead of having to add a new method.

PR #344, opened 2016-04-24
Move multires_image plug-in into multires_image package.

malban:
Move the multires image plug-in into multires_image package to simplify dependencies.

This should be merged after #343


malban:
Merged in with pull request #343

PR #357, opened 2016-05-18
Mapviz plug-in for calling the marti_nav_msgs::PlanRoute service (Jade)

malban:


malban:
Should be up to date now.

PR #359, opened 2016-05-25
Adds laps plugins

nick-alton:
It colors laps individually for easier inspection.


malban:
Is this basically a different draw style for the odometry message?

Can this be added as an display option for the current odometry plug-in?

pjreed:
Maybe it'd make sense to add a "Colorize Loops" checkbox to the odom plugin and add this functionality to it?

pjreed:
Abandoning this pull request because its features have been incorporated into #376.

PR #368, opened 2016-06-07
Fixing a crash in the PointCloud2 plugin (Indigo)

### LINKED ISSUE #316: PointCloud2 plugin occasionally crashes
The PC2 plugin sometimes causes mapviz to crash with a segmentation fault.  gdb indicates that the crash happens at:

```
#0  mapviz_plugins::PointCloud2Plugin::Draw (this=0x359db90, x=29.981998559053988, y=-576.14539349106178, scale=1.000000236541325) at /home/preed/src/DSAT_dev/src/MARTI/mapviz/mapviz_plugins/src/pointcloud2_plugin.cpp:578
#1  0x00007fb4043a88c6 in DrawPlugin (this=0x359db90, x=<optimized out>, y=<optimized out>, scale=<optimized out>) at /home/preed/src/DSAT_dev/src/MARTI/mapviz/mapviz/include/mapviz/mapviz_plugin.h:116
#2  mapviz::MapCanvas::paintEvent (this=0x3283a60, event=0x41efdb23) at /home/preed/src/DSAT_dev/src/MARTI/mapviz/mapviz/src/map_canvas.cpp:247
```

The code at pointcloud2_plugin.cpp:578 looks like:
```
    for (scan_it = scans_.begin(); scan_it != scans_.end(); scan_it++)
    {
      if (scan_it->transformed) // <-- this line
      {
```

I suspect the issue is because the member variable scans_ is modified by the PointCloud2Callback function; if that causes the object pointed to by scan_it to become invalid, it would likely cause a crash when trying to access it.

It would probably fix it to use a mutex to protect access to scans_.

pjreed:
Also sneaking in a few more changes:
- Caching transformed clouds to improve performance
- Properly saving the value of the "Color Transformer" combo box

Fixes #316

### DIFF
--- a/mapviz_plugins/src/pointcloud2_plugin.cpp
+++ b/mapviz_plugins/src/pointcloud2_plugin.cpp
@@ -93,9 +93,6 @@ namespace mapviz_plugins
     ui_.max_color->setColor(Qt::black);
     // Set color transformer choices
     ui_.color_transformer->addItem(QString("Flat Color"), QVariant(0));
-    //ui_.color_transformer->removeItem();
-
-
 
     QObject::connect(ui_.selecttopic,
                      SIGNAL(clicked()),
@@ -157,6 +154,10 @@ namespace mapviz_plugins
                                 const QColor &)),
                      this,
                      SLOT(DrawIcon()));
+    QObject::connect(this,
+                     SIGNAL(TargetFrameChanged(const std::string&)),
+                     this,
+                     SLOT(ResetTransformedPointClouds()));
 
     PrintInfo("Constructed PointCloud2Plugin");
   }
@@ -196,6 +197,15 @@ namespace mapviz_plugins
       painter.drawPoint(13, 2);
 
       icon_->SetPixmap(icon);
+    }
+  }
+
+  void PointCloud2Plugin::ResetTransformedPointClouds()
+  {
+    std::deque<Scan>::iterator scan_it = scans_.begin();
+    for (; scan_it != scans_.end(); ++scan_it)
+    {
+      scan_it->transformed = false;
     }
   }
 
@@ -240,7 +250,7 @@ namespace mapviz_plugins
     if (ui_.use_rainbow->isChecked())
     {  // Hue Interpolation
 
-      int hue = val * 255;
+      int hue = (int)(val * 255.0);
       return QColor::fromHsl(hue, 255, 127, 255);
     }
     else
@@ -249,9 +259,9 @@ namespace mapviz_plugins
       const QColor max_color = ui_.max_color->color();
       // RGB Interpolation
       int red, green, blue;
-      red = val * max_color.red() + ((1.0 - val) * min_color.red());
-      green = val * max_color.green() + ((1.0 - val) * min_color.green());
-      blue = val * max_color.blue() + ((1.0 - val) * min_color.blue());
+      red = (int)(val * max_color.red() + ((1.0 - val) * min_color.red()));
+      green = (int)(val * max_color.green() + ((1.0 - val) * min_color.green()));
+      blue = (int)(val * max_color.blue() + ((1.0 - val) * min_color.blue()));
       return QColor(red, green, blue, 255);
     }
   }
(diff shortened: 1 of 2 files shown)


mapviz_plugins/src/pointcloud2_plugin.cpp
```diff
@@ -271,13 +281,16 @@ namespace mapviz_plugins
...
-    for (; scan_it != scans_.end(); ++scan_it)
     {
-      std::vector<StampedPoint>::iterator point_it = scan_it->points.begin();
-      for (; point_it != scan_it->points.end(); point_it++)
+      QMutexLocker locker(&scan_mutex_);
+      std::deque<Scan>::iterator scan_it = scans_.begin();
+      for (; scan_it != scans_.end(); ++scan_it)
       {
-        point_it->color = CalculateColor(*point_it);
+        std::deque<StampedPoint>::iterator point_it = scan_it->points.begin();
```
evenator: Is there a reason these iterators are outside of the `for` block scope?

mapviz_plugins/src/pointcloud2_plugin.cpp
```diff
@@ -333,10 +349,11 @@ namespace mapviz_plugins
 
   void PointCloud2Plugin::BufferSizeChanged(int value)
   {
-    buffer_size_ = value;
+    buffer_size_ = (size_t)value;
```
evenator: See comment above about `static_cast`

mapviz_plugins/src/pointcloud2_plugin.cpp
pjreed: I believe @nick-alton wrote this originally, so probably a question for him.  I'd guess it's just to make the `for` line shorter.

mapviz_plugins/src/pointcloud2_plugin.cpp
```diff
@@ -428,16 +444,20 @@ namespace mapviz_plugins
...
+          if (field == saved_color_transformer_) {
+            // The very first time we see a new set of features, that means the
+            // plugin was just created; if we have a saved value, set the current
+            // index to that and clear the saved value.
+            new_feature_index = label;
+            saved_color_transformer_ = "";
           }
-          ui_.color_transformer->addItem(QString(a), QVariant(label));
+
+          ui_.color_transformer->addItem(QString::fromStdString(field), QVariant(label));
```
evenator: :+1:  This change.

mapviz_plugins/src/pointcloud2_plugin.cpp
pjreed: I don't see an above comment?

But I'm guessing you commented that I was lazy for using a C-style cast, and you are correct, it just slipped my mind that I should really use `static_cast` instead.

mapviz_plugins/src/pointcloud2_plugin.cpp
```diff
@@ -494,7 +517,7 @@ namespace mapviz_plugins
       case 1:
         return *reinterpret_cast<const int8_t*>(data + feature_info.offset);
       case 2:
-        return *reinterpret_cast<const uint8_t*>(data + feature_info.offset);
+        return *(data + feature_info.offset);
       case 3:
         return *reinterpret_cast<const int16_t*>(data + feature_info.offset);
```
evenator: Why are the return types inconsistent here?

pjreed: `reinterpret_cast<const uint8_t*>` is redundant; `data + feature_info.offset` is a `const uint8_t*`.

mapviz_plugins/src/pointcloud2_plugin.cpp
```diff
@@ -654,28 +681,35 @@ namespace mapviz_plugins
...
       use_latest_transforms_ = false;
-      if (GetTransform(scan.source_frame, scan.stamp, transform))
+      for (; scan_it != scans_.end(); ++scan_it)
       {
-        scan.transformed = true;
-        std::vector<StampedPoint>::iterator point_it = scan.points.begin();
-        for (; point_it != scan.points.end(); ++point_it)
+        Scan& scan = *scan_it;
+
+        if (!scan_it->transformed)
```
evenator: Does caching the transformed points have any possible nasty side effects? I'm asking because I honestly don't know, and caching makes me nervous.

pjreed: Answered out loud by @malban.  In this particular case, since the PointCloud2 plugin doesn't use the latest transforms and the source frame for the stored points won't change, your transform should only change if the target frame changes, and the cache will be invalidated if the target frame changes.

evenator:
LGTM

At some point, I'll look into doing a style review on this repo, because there are some parts of it that are a mess.

PR #370, opened 2016-06-08
Implementing support for the ARROW marker type (Indigo)

### LINKED ISSUE #365: Cannot display Arrow marker
Hello!

I have trouble visualizing an arrow marker.

I think I have initialized everything properly, but I can't get it to work, even though I get an OK message in the topic (**marker/attractor**) subscription from Mapviz.

You can see the topic content and the Mapviz window in the attached picture.
screenshot from 2016-06-06 11 57 12

fcu is the same frame as the rover, basically the body frame.

If I swap the marker type to 1 (a cube) then it displays correctly, which is even weirder.

I have tried

- passing all-zero pose and a 2-long points array
- passing only pose data
- both simultaneously

Can you replicate my issue or does it work for you?

Thank you in advance.

pjreed:
Resolves #365

### DIFF
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -324,7 +324,8 @@ namespace mapviz_plugins
       arrowOffset = tf::Point(0.25 * markerData.scale_y, 0.0, 0.0);
     }
 
-    tfScalar angle = tf::Point(1.0, 0.0, 0.0).angle(point.transformed_arrow_point - point.transformed_point);
+    tf::Vector3 pointDiff = point.transformed_arrow_point - point.transformed_point;
+    double angle = std::atan2(pointDiff.getY(), pointDiff.getX());
 
     tf::Transform left_tf(tf::createQuaternionFromRPY(0, 0, M_PI*0.75 + angle));
     tf::Transform right_tf(tf::createQuaternionFromRPY(0, 0, -M_PI*0.75 + angle));
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -287,7 +287,7 @@ namespace mapviz_plugins
   }
 
   /**
-   * Gives a MarkerData that represents an arrow and a transform, this function
+   * Given a MarkerData that represents an arrow and a transform, this function
    * will generate the points involved in drawing the arrow and then transform
    * all of them into the target frame.
    * @param[inout] markerData A marker that represents an arrow.
@@ -296,7 +296,7 @@ namespace mapviz_plugins
   void MarkerPlugin::transformArrow(MarkerData& markerData,
                       const swri_transform_util::Transform& transform)
   {
-    // The firstpoint in the markerData.points array always represents the
+    // The first point in the markerData.points array always represents the
     // base of the arrow.
     StampedPoint& point = markerData.points.front();
     tf::Point arrowOffset;
@@ -320,7 +320,7 @@ namespace mapviz_plugins
       // length and angle are determined by them.
       point.transformed_point = transform * point.point;
       point.transformed_arrow_point = transform * point.arrow_point;
-      // Also, in those mode, scale_y is the diameter of the arrow's head.
+      // Also, in this mode, scale_y is the diameter of the arrow's head.
       arrowOffset = tf::Point(0.25 * markerData.scale_y, 0.0, 0.0);
     }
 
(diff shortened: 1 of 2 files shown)


mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -254,6 +286,53 @@ namespace mapviz_plugins
...
+      // If the markerData has two points, that means that the start and end points
+      // of the arrow were explicitly specified in the original message, so the
+      // length and angle are determined by them.
+      point.transformed_point = transform * point.point;
+      point.transformed_arrow_point = transform * point.arrow_point;
+      // Also, in those mode, scale_y is the diameter of the arrow's head.
+      arrowOffset = tf::Point(0.25 * markerData.scale_y, 0.0, 0.0);
+    }
+
+    tfScalar angle = tf::Point(1.0, 0.0, 0.0).angle(point.transformed_arrow_point - point.transformed_point);
```
pjreed: This seems kind of ugly to me, but I didn't find an obviously better way to find the angle of a line that passes through both points; tf::Point::angle returns the difference between the angles of two points relative to the origin.  Am I missing the right way to do this?

mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -328,44 +407,90 @@ namespace mapviz_plugins
         {
           if (marker.transformed)
           {
-            glColor4f(marker.color.redF(), marker.color.greenF(), marker.color.blueF(), 1.0f);
+            glColor4d(marker.color.redF(), marker.color.greenF(), marker.color.blueF(), marker.color.alphaF());
```
pjreed: Not exactly related to this change, but I the alpha channel for markers wasn't being respected here.

mapviz_plugins/src/marker_plugin.cpp
pjreed: Decided to use std::atan2 instead to get the angle of the line through those points.  It seems a little bit simpler.

Georacer:
It works!
screenshot from 2016-06-09 08 44 01
Is this going on indigo-devel soon?

pjreed:
Great, thanks for testing it!

It might be a little bit before it's actually in the ROS repository; we'll need to publish a new release of mapviz (and first decide if there are any other fixes or features we want for the next release), and then we'll have to wait until the next time ROS releases a new version of Indigo, which can be a bit intermittent.

Georacer:
Okay, no worries. Right now I have checked out your PR and I'm at the 
end of this project anyway, so I don't need to include anymore features 
of mapviz. Thanks for letting me know. Mapviz proved a really useful 
tool so far.

mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -203,7 +203,39 @@ namespace mapviz_plugins
...
       }
 
-      if (markerData.display_type == visualization_msgs::Marker::CYLINDER ||
+      if (markerData.display_type == visualization_msgs::Marker::ARROW)
+      {
+        StampedPoint point;
+        point.color = markerData.color;
+        point.orientation = orientation;
+
+        if (marker.points.empty())
```
evenator: Change this check to size() < 2 to prevent a segfault in the else on malformed messages with exactly 1 point.

mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -203,7 +203,39 @@ namespace mapviz_plugins
...
+        {
+          // Otherwise the "points" array should have exactly two values, the
+          // start and end of the arrow.
+          point.point = tf::Point(marker.points[0].x, marker.points[0].y, marker.points[0].z);
+          point.arrow_point = tf::Point(marker.points[1].x, marker.points[1].y, marker.points[1].z);
+        }
+
+        markerData.points.push_back(point);
+
+        if (!marker.points.empty()) {
```
evenator: This check will also have to be changed if the one above is changed.

mapviz_plugins/src/marker_plugin.cpp
pjreed: If a malformed message with exactly 1 point arrives, I think it might be better to just drop it entirely and log the error rather than try to continue on as thought it was a valid message with 0 points.

evenator: That works too. I just don't want to crash mapviz if someone sends it a bad message.

evenator:
LGTM. @malban?

malban:
:+1:

PR #374, opened 2016-06-10
Ensuring that Mapviz won't subscribe to empty topic names (Jade)

### LINKED ISSUE #327: Clearing Topic Name Causes Unhandled Exception Crash
If a plugin does not check if a topic name is blank before subscribing to it, it causes an error if more than one plugin is subscribed to the blank topic because ROS throws an exception that the topic types don't match. I encountered this bug when I cleared the topic name on an image plugin and then on a disparity plugin:

```
Qt has caught an exception thrown from an event handler. Throwing
exceptions from an event handler is not supported in Qt. You must
reimplement QApplication::notify() and catch all exceptions there.

terminate called after throwing an instance of 'ros::ConflictingSubscriptionException'
  what():  Tried to subscribe to a topic with the same name but different md5sum as a topic that was already subscribed [stereo_msgs/DisparityImage/04a177815f75271039fa21f16acad8c9 vs. sensor_msgs/Image/060021388200f6f0f447d0fcd9c64743]
```

Plugins known to have this bug (check off when fixed):

- [ ] image
- [ ] disparity

pjreed:
I cleaned up and made more consistent the code for handling subscriptions for all topics.
The behavior is now:
- All input is trimmed before processing
- If a topic name is empty, the old subscriber will be shut down and will not subscribe to the empty topic

I also cleaned up and reformatted the code in the Attitude Indicator plugin because it was pretty messy.

Resolves #327

### DIFF
--- a/mapviz_plugins/src/attitude_indicator_plugin.cpp
+++ b/mapviz_plugins/src/attitude_indicator_plugin.cpp
@@ -103,18 +103,22 @@ AttitudeIndicatorPlugin::~AttitudeIndicatorPlugin()
 
  void AttitudeIndicatorPlugin::TopicEdited()
  {
-     if (ui_.topic->text().toStdString() != topic_)
+     std::string topic = ui_.topic->text().trimmed().toStdString();
+     if (topic != topic_)
      {
        initialized_ = true;
        has_message_ = false;
-       topic_ = boost::trim_copy(ui_.topic->text().toStdString());
        PrintWarning("No messages received.");
 
        odometry_sub_.shutdown();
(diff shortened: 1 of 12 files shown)


evenator:
I hate to be *that guy*, but can you split the code cleanup into a separate PR? It's making it hard for me to read the diffs.

pjreed:
Sigh, I suppose that's reasonable.  I'll make a different branch for the cleanup; I've reverted it out of these two pull requests.

evenator:
LGTM, but I don't think our test coverage is good enough to prove anything.

pjreed:
Test coverage?  What test coverage?

PR #376, opened 2016-06-21
Point drawing abstraction

nick-alton:
Updates gps,navsat,odometry,path, and tf_frame plugins to use the
abstract point drawing class. Also adds the draw laps functionality
which will change the color of the path as it passes a base point for
ease of visibility, currently implemented on gps and odometry plugins.


mapviz_plugins/include/mapviz_plugins/point_draw.h
```diff
@@ -0,0 +1,115 @@
...
+#include <QWidget>
+
+// ROS libraries
+#include <ros/ros.h>
+#include <tf/transform_datatypes.h>
+
+#include <mapviz/map_canvas.h>
+namespace mapviz_plugins
+{
+  class PointDraw : public mapviz::MapvizPlugin
```
pjreed: I'm not sure I like the name "PointDraw" for a class, probably because it's not a noun.  Why not something like "PointDrawingPlugin"?

pjreed:
As a general comment, it looks like there are several files in this pull request that got re-indented and reformatted but are unrelated to the actual changes.  We could probably stand to do a reformatting pass over all of the plugins just to make sure they look pretty, but I would rather have that in a separate PR and have this one only contain the path drawing-related changes.

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+        glVertex2f(it->transformed_arrow_point.getX(),
+                   it->transformed_arrow_point.getY());
+        glVertex2f(it->transformed_arrow_left.getX(),
+                   it->transformed_arrow_left.getY());
+
+        glVertex2f(it->transformed_arrow_point.getX(),
+                   it->transformed_arrow_point.getY());
+        glVertex2f(it->transformed_arrow_right.getX(),
+                   it->transformed_arrow_right.getY());
```
pjreed: I see some copy/pasted blocks of code in here that could be reduced down to a single function.  Maybe turn the block above here into a function that looks something like `void DrawArray(const StampedPoint& point)`...

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+      glVertex2f(cur_point_.transformed_arrow_point.getX(),
+                 cur_point_.transformed_arrow_point.getY());
+      glVertex2f(cur_point_.transformed_arrow_left.getX(),
+                 cur_point_.transformed_arrow_left.getY());
+
+      glVertex2f(cur_point_.transformed_arrow_point.getX(),
+                 cur_point_.transformed_arrow_point.getY());
+      glVertex2f(cur_point_.transformed_arrow_right.getX(),
+                 cur_point_.transformed_arrow_right.getY());
```
pjreed: And here

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+            glVertex2f(it->transformed_arrow_point.getX(),
+                       it->transformed_arrow_point.getY());
+            glVertex2f(it->transformed_arrow_left.getX(),
+                       it->transformed_arrow_left.getY());
+
+            glVertex2f(it->transformed_arrow_point.getX(),
+                       it->transformed_arrow_point.getY());
+            glVertex2f(it->transformed_arrow_right.getX(),
+                       it->transformed_arrow_right.getY());
```
pjreed: And here

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+
+          glVertex2f(it->transformed_arrow_point.getX(),
+                     it->transformed_arrow_point.getY());
+          glVertex2f(it->transformed_arrow_left.getX(),
+                     it->transformed_arrow_left.getY());
+
+          glVertex2f(it->transformed_arrow_point.getX(),
+                     it->transformed_arrow_point.getY());
+          glVertex2f(it->transformed_arrow_right.getX(),
+                     it->transformed_arrow_right.getY());
```
pjreed: And here

mapviz_plugins/src/gps_plugin.cpp
```diff
@@ -185,19 +126,23 @@ namespace mapviz_plugins
...
     // Furthermore, the track rotates in the opposite direction and is also
     // offset by 90 degrees, so all of that has to be compensated for.
-    stamped_point.orientation = tf::createQuaternionFromYaw((-gps->track * (M_PI / 180.0)) + M_PI_2);
+    stamped_point.orientation =
+        tf::createQuaternionFromYaw((-gps->track * (M_PI / 180.0)) + M_PI_2);
 
-    if (points_.empty() || stamped_point.point.distance(points_.back().point) >= position_tolerance_)
+    if (points_.empty() ||
+        stamped_point.point.distance(points_.back().point) >=
+            position_tolerance_)
```
jatowler: In compound conditionals like this, especially wrapped over multiple lines, adding some clarifying parentheses would help:
```
if (points_.empty() ||
    (stamped_point.point.distance(points_.back().point) >= position_tolerance_))
```

mapviz_plugins/src/gps_plugin.cpp
```diff
@@ -39,17 +39,12 @@
 
 // Declare plugin
 #include <pluginlib/class_list_macros.h>
-PLUGINLIB_DECLARE_CLASS(
-    mapviz_plugins,
-    gps,
-    mapviz_plugins::GpsPlugin,
-    mapviz::MapvizPlugin);
+PLUGINLIB_DECLARE_CLASS(mapviz_plugins, gps, mapviz_plugins::GpsPlugin,
+                        mapviz::MapvizPlugin);
```
jatowler: This change makes the arguments to this macro much harder to read. If it's the indentation you don't like, then restyling like this would also work:
```
PLUGINLIB_DECLARE_CLASS(mapviz_plugins,
                        gps,
                        mapviz_plugins::GpsPlugin,
                        mapviz::MapvizPlugin);
```
Every `_plugin.cpp` has this call, and they should be consistent, so if you fix one, make sure to fix them all.

mapviz_plugins/src/navsat_plugin.cpp
```diff
@@ -418,9 +293,9 @@ namespace mapviz_plugins
     std::string draw_style = ui_.drawstyle->currentText().toStdString();
     emitter << YAML::Key << "draw_style" << YAML::Value << draw_style;
 
-    emitter << YAML::Key << "position_tolerance" << YAML::Value << position_tolerance_;
+    emitter << YAML::Key << "position_tolerance" << YAML::Value
+            << position_tolerance_;
```
jatowler: This line, now that it's wrapped, would gloss better like this:
```
emitter << YAML::Key   << "position_tolerance" <<
           YAML::Value << position_tolerance_;
```

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -198,18 +141,17 @@ namespace mapviz_plugins
...
+                                    odometry->pose.pose.position.y,
+                                    odometry->pose.pose.position.z);
 
     stamped_point.orientation = tf::Quaternion(
-        odometry->pose.pose.orientation.x,
-        odometry->pose.pose.orientation.y,
-        odometry->pose.pose.orientation.z,
-        odometry->pose.pose.orientation.w);
+        odometry->pose.pose.orientation.x, odometry->pose.pose.orientation.y,
+        odometry->pose.pose.orientation.z, odometry->pose.pose.orientation.w);
```
jatowler: This re-formatting makes this call significantly harder to read. Once an argument list wraps onto multiple lines, it is almost always preferable to split every argument onto its own line.

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+  {
+    bool transformed = false;
+    glColor4f(color_.redF(), color_.greenF(), color_.blueF(), 0.5);
+    glLineWidth(3);
+    QColor base_color = color_;
+    if (laps_.size() != 0)
+    {
+      for (int i = 0; i < laps_.size(); i++)
+      {
+        int hue = color_.hue() + (i + 1) * 10 * 3.1415926;
```
jatowler: This line has two [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)). One of them looks like an approximation for π, for which you should use `swri_math_util::_pi`. The other is inscrutable and should be assigned to a named constant immediately above this line to explain its purpose.

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,485 @@
...
+  {
+    bool transformed = false;
+    glColor4f(color_.redF(), color_.greenF(), color_.blueF(), 0.5);
+    glLineWidth(2);
+    QColor base_color = color_;
+    if (laps_.size() != 0)
+    {
+      for (int i = 0; i < laps_.size(); i++)
+      {
+        int hue = color_.hue() + (i + 1) * 10 * 3.1415926;
```
jatowler: See earlier comment about magic numbers. If this is a common operation, it should be extracted into a function with an explanatory name.

mapviz_plugins/src/tf_frame_plugin.cpp
```diff
@@ -462,17 +266,20 @@ namespace mapviz_plugins
...
-  void TfFramePlugin::SaveConfig(YAML::Emitter& emitter, const std::string& path)
+  void TfFramePlugin::SaveConfig(YAML::Emitter& emitter,
+                                 const std::string& path)
   {
-    emitter << YAML::Key << "frame" << YAML::Value << ui_.frame->text().toStdString();    
-    emitter << YAML::Key << "color" << YAML::Value << ui_.color->color().name().toStdString();
+    emitter << YAML::Key << "frame" << YAML::Value
+            << ui_.frame->text().toStdString();
+    emitter << YAML::Key << "color" << YAML::Value
+            << ui_.color->color().name().toStdString();
```
jatowler: Once again, the preferred wrapping for these lines looks like
```
emitter << YAML::Key   << key <<
           YAML::Value << value;
```
or
```
emitter << YAML::Key   << key
        << YAML::Value << value;
```

mapviz_plugins/src/tf_frame_plugin.cpp
```diff
@@ -462,17 +266,20 @@ namespace mapviz_plugins
...
 
     std::string draw_style = ui_.drawstyle->currentText().toStdString();
     emitter << YAML::Key << "draw_style" << YAML::Value << draw_style;
 
-    emitter << YAML::Key << "position_tolerance" << YAML::Value << position_tolerance_;
+    emitter << YAML::Key << "position_tolerance" << YAML::Value
+            << position_tolerance_;
```
jatowler: See above about wrapping these lines

mapviz_plugins/src/point_drawing_plugin.cpp
evenator: I'd recommend `M_PI` over `swri_math_util::_pi` because swri_math_util is not currently a dependency of mapviz, and it seems unnecessary to bring it in just for a constant.

jatowler: Agreed; I just assumed it was being brought in somewhere. If it's not, `M_PI` is better.

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,437 @@
...
+                 cur_point_.transformed_point.getY());
+
+      transformed = true;
+    }
+
+    glEnd();
+
+    return transformed;
+  }
+  bool PointDrawingPlugin::DrawArray(const StampedPoint& it)
```
pjreed: Let's rename this to DrawArrow.

pjreed:
When testing it, it seems like the "Draw Laps" checkbox clears the buffer when it is checked.  Is this intentional?  It seems like somebody might want to highlight laps on a track that they've already recorded.

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -0,0 +1,437 @@
...
+#include <swri_transform_util/transform_util.h>
+
+namespace mapviz_plugins
+{
+  PointDrawingPlugin::PointDrawingPlugin()
+      : new_lap_(true),
+        got_begin_(false),
+        draw_style_(LINES),
+        position_tolerance_(0.0),
+        lap_checked_(false)
```
pjreed: This class has a few primitive types that you are not initializing.  It would probably be a good idea to also set:
```
int buffer_size_;
bool covariance_checked_;
int buffer_holder_
```

Also make sure they're initialized here in the same order that they're declared in the .h file.

pjreed:
Aside from the uninitialized member variables, this looks good to me.  Rebase it down to one commit and I'll merge it in.

pjreed:
I managed to make a big mess out of this rebase, so I'm just going to cherry-pick the individual files and make a new PR at #382.

PR #399, opened 2016-07-08
Adding a dialog for selecting services by type (Kinetic)

pjreed:
This dialog is similar to the ones for listing topics or TF frames, but it is
a little different under the hood.  Notably:
- It relies on the rosapi node in order to be able to search for services
- Since searching is done via a service call, ROS communication is handled
  on a separate thread that will not block the GUI
- Unlike topics, only searching for a single service type is supported

Conflicts:
	mapviz/package.xml

### DIFF
--- a/mapviz/CMakeLists.txt
+++ b/mapviz/CMakeLists.txt
@@ -4,6 +4,7 @@ project(mapviz)
 find_package(catkin REQUIRED COMPONENTS
   marti_common_msgs
   message_generation
+  rosapi
   roscpp
   #rqt_gui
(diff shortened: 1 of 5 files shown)


pjreed:
Apparently `rosapi` hasn't been released into kinetic yet.  Here's the appropriate ticket: https://github.com/RobotWebTools/rosbridge_suite/issues/214

I suppose this PR will have to wait until then.

evenator:
rosbridge is up in shadow-fixed for kinetic.

PR #400, opened 2016-07-11
Fix compiler warning in mapviz widgets.h

evenator:
QListWidget::count() returns an int, so comparing it to a size_t causes a sign-compare warning.


pjreed:
:+1:

PR #416, opened 2016-08-10
Add option to not scale arrows with zoom level (Indigo)

### LINKED ISSUE #414: Scaling of GPS/NavSatFix markers
We have a use case where we want to see multiple markers at once on the map, which requires us to be zoomed out far enough that the arrows that are drawn are too small to see the detail. It would be nice if there was another UI field to adjust the size of the drawn markers.

pjreed:
This adds a checkbox to all of the plugins that can draw a series of
coordinates as arrows; i. e., the GPS, NavSatFix, Odometry, and TF Frame
plugins.  This checkbox will control whether the arrows are drawn at a fixed
size regardless of zoom level or whether they are scaled with the zoom level.

Resolves #414

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/point_drawing_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/point_drawing_plugin.h
@@ -95,8 +95,10 @@ namespace mapviz_plugins
     virtual void DrawIcon();
     virtual void SetDrawStyle(QString style);
     virtual void SetScaledArrows(bool isChecked);
+    virtual void SetArrowSize(int arrowSize);
 
    protected:
+    int arrow_size_;
     DrawStyle draw_style_;
     StampedPoint cur_point_;
     std::list<StampedPoint> points_;
--- a/mapviz_plugins/include/mapviz_plugins/point_drawing_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/point_drawing_plugin.h
@@ -81,7 +81,7 @@ namespace mapviz_plugins
     {
     }
     virtual void Transform();
-    virtual bool DrawPoints();
+    virtual bool DrawPoints(double scale);
     virtual bool DrawArrows();
     virtual bool DrawArrow(const StampedPoint& point);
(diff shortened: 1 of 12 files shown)


jhoare:
Thanks for this! 

One thing I noticed while running it: It appears the "Scaled Arrows" checkbox is inverted from what would be expected. The arrows will change with zoom level when the box is unchecked, and stays static when it is checked. 

It would also be nice to be able to manually set the scaling factor by some multiplier, e.g be able to change the arrows so they draw 2x or 1.5x larger than they are being drawn now. When I am giving a demo where people are looking at a TV or projector I may want the arrows to be larger than when I am doing devleopment/debugging work. 

The arrows may also look smaller for me since I am using a 4k display and UI elements being small is a commonly experienced issue.

pjreed:
Ah, that's a good point.  They look just right for me on a 1920x1080 display, so they would probably be far too small on a 4k display. :smile:  I'll add a way to adjust the size.

pjreed:
I've added a slider that can be used to adjust the size of the arrows.  For the "Scaled arrows:" label, I think that might be a bit confusing; my line of thought was that it would cause the arrows to be scaled with respect to the zoom level, so naturally they would change size as you zoomed in and out, but it's also perfectly reasonable to think that they should be scaled with respect to the user's display instead and so would not change with the zoom level.  I'm trying to think of a clearer label but haven't had any luck so far...

jhoare:
Thanks, I've updated and the slider bar works great!

After thinking about the label you're right, its easy to look at it from both perspectives as to what "scaling" the arrows entails. (e.g. relative to the background or relative to the window view) 

Either way, I appreciate the added feature we will certainly make use of it!

jatowler:
Not having looked at it, would "scale arrows with zoom" or "scale arrows with frame" or "zoom arrows with map" or something along those lines work (or fit in the space)? Or, the opposite, "Static arrow size"?

pjreed:
I think I'll go with "Static Arrow Sizes".  That fits and is hopefully a little clearer.

Issue #421, opened 2016-08-24
tile_map: Maps not reliably loading 


jhoare:
When attempting to load mapviz using a tile_map, it does not load every square of the map. Sometimes all (or at least most) of the squares will load, but most of the time it is very patchy and it seems like it has given up loading. 

I do sometimes get a "Too Many Failures.." message to the console, but I often don't get any errors printed to the console. 

I'm mostly working with the Bing Maps source but i seem to get similar results with the stamen sources as well. 

I wonder if this too is related to my 4k display, and so maybe many more tiles are being requested than you guys work with?

pjreed:
I think this is the same as #342.

I don't think this is specifically because of a 4k display, but it may exacerbate the problem.  Basically, when it tries to load a lot of tiles at once, some of the HTTP requests may time out or fail, and currently the `tile_map` plugin doesn't retry failed connections; it really should.

pjreed:
Closing since this was fixed in #423.

PR #423, opened 2016-08-26
Rewrite tile_map loading to be more reliable

### LINKED ISSUE #342: MapQuest tiles not loading
Occasionally some tiles are not loaded/displayed properly (see screenshot). While i understand that this is most likely a problem with MapQuest i think there should be a mechanism to retry at least a few times after some interval. A restart of mapviz/the tile_map plugin might sovle the problem for these tiles but missing tiles like this seem to be a common occurence and you can't rely on restarting until all tiles in your area are loaded.

screenshot from 2016-04-21 11 39 09

pjreed:
This changes how the tile_map plugin handles making network requires for tiles.
It will now:
- Use thread conditions to prompt loading rather than spinning
- Prioritize loading tiles in the visible area
- Retry a failed network request up to 5 times
- Not discard tile requests if there are more than 100 in the queue

These changes should significantly reduce (if not completely eliminate) the
number of tiles that fail to load and hopefully make tiles within the visible
area appear faster when there are many in the queue.

Fixes #342 and #421.

### DIFF
--- a/tile_map/include/tile_map/image_cache.h
+++ b/tile_map/include/tile_map/image_cache.h
@@ -31,6 +31,7 @@
 #define TILE_MAP_IMAGE_CACHE_H_
 
 #include <string>
+#include <limits>
 
 #include <boost/cstdint.hpp>
 #include <boost/shared_ptr.hpp>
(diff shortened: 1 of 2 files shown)


jhoare:
Thanks! Anecdotally, it seems like it takes a while to start displaying images, but playing around it looks like everything does load, which is a big improvement!

pjreed:
Yeah, it does seem like it can take a little longer to initially display tiles.  It could probably stand some more improvement, but this is enough better that I'll go ahead and merge it in.

PR #432, opened 2016-10-03
Add a GUI for controlling the Image Transport (Indigo)

### LINKED ISSUE #430: GUI support of image_transport compressed datatypes
We would like to use the compressed image formats so we can more efficiently stream images over a wireless network.

pjreed:
This will add a sub-menu under the "View" menu that will:
- List all available image transports
- Indicate which one is currently the default
- Allow the user to choose which one will be used for new ImageTransport subscriptions
- Save and restore this setting to Mapviz's config file
- Cause any `image` plugins using the default transport to resubscribe

In addition, the `image` plugin now has a menu that can be used to change the transport for that specific plugin so that it is different from the default.

Fixes #430

### DIFF
--- a/mapviz/src/mapviz.cpp
+++ b/mapviz/src/mapviz.cpp
@@ -65,10 +65,13 @@
 #include <mapviz/config_item.h>
 #include <QtGui/QtGui>
 
+#include <image_transport/image_transport.h>
+
 namespace mapviz
 {
 const QString Mapviz::ROS_WORKSPACE_VAR = "ROS_WORKSPACE";
 const QString Mapviz::MAPVIZ_CONFIG_FILE = "/.mapviz_config";
+const std::string Mapviz::IMAGE_TRANSPORT_PARAM = "image_transport";
 
 Mapviz::Mapviz(bool is_standalone, int argc, char** argv, QWidget *parent, Qt::WFlags flags) :
     QMainWindow(parent, flags),
@@ -167,6 +170,11 @@ Mapviz::Mapviz(bool is_standalone, int argc, char** argv, QWidget *parent, Qt::W
   connect(stop_button_, SIGNAL(clicked()), this, SLOT(StopRecord()));
   connect(screenshot_button_, SIGNAL(clicked()), this, SLOT(Screenshot()));
 
+  image_transport_menu_ = new QMenu("Default Image Transport", ui_.menu_View);
+  ui_.menu_View->addMenu(image_transport_menu_);
+
+  connect(image_transport_menu_, SIGNAL(aboutToShow()), this, SLOT(UpdateImageTransportMenu()));
+
   ui_.bg_color->setColor(background_);
   canvas_->SetBackground(background_);
 }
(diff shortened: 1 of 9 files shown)


pjreed:
It would be nice if we could set this on a per-topic setting, but the `image_transport` param globally affects all ImageTransport subscribers.  Not much we can do about that.

But before we merge this in, something to consider: should existing ImageTransport subscriptions automatically re-subscribe using the new transport if it's changed?

This would require adding a mechanism to notify plugins that they should re-subscribe to their topics.  We could probably just add a slot to the MapvizPlugin class and notify it when the transport is changed.  Image plugins could respond to it by re-subscribing, and other plugins could just ignore it.

On the other hand, is this something we actually want to do?  It' d be weird, but somebody could conceivably want to subscribe to different topics using different transports...

mapviz/src/mapviz.cpp
```diff
@@ -201,7 +209,23 @@ void Mapviz::Initialize()
       connect(&spin_timer_, SIGNAL(timeout()), this, SLOT(SpinOnce()));
     }
 
-    node_ = new ros::NodeHandle("mapviz");
+    node_ = new ros::NodeHandle("~");
```
pjreed: Explicitly setting this to `mapviz` would cause the node handle to use "mapviz" as its name even if the actual mapviz node had a different name, which it normally would since Mapviz's node name is anonymized.  This would make it rather difficult to set/get params directly on this node.  Setting it to `~` instead will make it use the node's real name.

evenator:
Could you create a private sub-namespace nodehandle for each image subscription to allow per-topic image transport type? Or is there maybe a way to access the image transport subscriber at a lower level than the magic subscriber interface?

pjreed:
Hm... it might be possible to do so by having each instance of the Image plugin create an instance of the ImageTransport class that has a sub-namespaced node handle that's unique to that instance (see [image_transport::ImageTransport()](http://docs.ros.org/api/image_transport/html/classimage__transport_1_1TransportHints.html#a1685aa1484249c2a4f70d91a2eac470a)) and use the parameter on that to control the transport.  I'll have to experiment with it.

mapviz/src/mapviz.cpp
evenator: > But before we merge this in, something to consider: should existing ImageTransport subscriptions automatically re-subscribe using the new transport if it's changed?

>This would require adding a mechanism to notify plugins that they should re-subscribe to their topics.

I think it should automatically resubscribe, and I can imagine there will be future use-cases for that interface.

evenator:
> But before we merge this in, something to consider: should existing ImageTransport subscriptions automatically re-subscribe using the new transport if it's changed?

>This would require adding a mechanism to notify plugins that they should re-subscribe to their topics.

I think it should automatically resubscribe, and I can imagine there will be future use-cases for that interface.

pjreed:
I decided against adding it for every plugin because not every plugin needs to re-subscribe when the default image transport changes.  The solution I went with is a little kludgey; when creating plugins, it specifically looks for "mapviz_plugins/image" plugins and hooks up a signal generated by the transport changing to a slot that will make them resubscribe.

mapviz/src/mapviz.ui
```diff
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>732</width>
+    <width>600</width>
```
evenator: Unrelated change?

mapviz_plugins/src/image_config.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>400</width>
-    <height>300</height>
+    <width>334</width>
```
evenator: Another unrelated change?

mapviz/src/mapviz.ui
pjreed: Added in inadvertently due to using Qt Designer to edit the .ui file.  It shouldn't actually affect anything.

PR #447, opened 2016-12-09
Initialize pointcloud2 buffer size variable. (Kinetic)

malban:
Initialize pointcloud2 buffer size variable.


malban:
@pjreed Do you know why this is failing to build?

evenator:
@malban There's an opencv problem in marti_common. Should be fixed shortly.

PR #457, opened 2016-12-21
Add parameter to disable auto-saving functionality.

malban:
Add parameter to disable auto-saving functionality.


pjreed:
Would it make sense to also add a menu option to disable it in the GUI?  (although then the question is "should you auto-save changing the state of the auto-save switch" ...)

malban:
That might be useful, but someone can make a different pull request if they really want it.

PR #468, opened 2016-12-22
Add plug-in for drawing and publishing a polygon.

malban:
This plug-in allows you to draw a polygon on the screen and publish it to a topic.

draw_polygon


pjreed:
:+1:

Might be nice to be able to set the z-value on the polygon, although it doesn't matter for mapviz and I don't know if anybody else cares.

PR #470, opened 2017-01-18
added left_offset to robot image plugin config

nealseegmiller:
previously the plugin assumed vehicle frame origin to be located
at the vehicle center. This allows the user to specify the distance
from the left of image (rear of vehicle) to the origin.

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/robot_image_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/robot_image_plugin.h
@@ -83,13 +83,16 @@ namespace mapviz_plugins
     void FrameEdited();
     void WidthChanged(double value);
     void HeightChanged(double value);
+    void LeftOffsetChanged(double value);
 
   private:
     Ui::robot_image_config ui_;
     QWidget* config_widget_;
 
-    double width_;
-    double height_;
+    double width_; //image width, if vehicle frame is x-forward this corresponds to vehicle length
+    double height_; //image height, corresponds to vehicle width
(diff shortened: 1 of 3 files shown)


malban:
Wouldn't it be better to either:
1.  Add a static transform in TF to account for the offset
2.  Add fields for offset_x and offset_y as offsets from the center of the image

evenator:
+1 for offsets from image center.

Requiring a TF seems like more work than I want to do most of the time.

nealseegmiller:
Good suggestions. My thoughts are:
1. Always having to add a static transform to a vehicle_center frame in the
launch files would be cumbersome. The vehicle center frame would only be
useful for this plugin so it makes sense to do the work in the plugin
2. The reason I used left_offset is that it corresponds directly to a
vehicle dimensions parameter - it is equals the "rear overhang", or
distance from the rear of the vehicle to the rear axle. You can easily look
up rear overhang in the specs for commercial vehicles. You can't look up
the offset from vehicle center directly, you'd have to mentally compute it.
I didn't include an offset for y because I've never needed it personally.
The origin has always been at the image center. It wouldn't hurt to provide
one though.

malban:
Even so, I still prefer offsetting from the center.  The math for calculating the offset should be trivial given the overhang and vehicle length.

nealseegmiller:
Okay, it's a little less convenient for my use case but seems cleaner and
more general. I'll change it.

nealseegmiller:
I removed the left_offset parameter and added offset_x, offset_y instead. I have verified it works on my computer.

PR #485, opened 2017-05-11
Add /wgs84 frame to point click publisher when available.

malban:
Add /wgs84 frame to point click publisher when available.


pjreed:
:+1: 
Looks good, works for me.

PR #490, opened 2017-07-19
 Added timestamp display to odometry and fixed color of covariance display

### LINKED ISSUE #488: Odometry covariance ellipse is always drawn in green
Rather than the color of the odometry trace.

See for example: 

mapviz_20170616t163507_476232

matt-attack:
Changed color of covariance display to match that of the odometry. Fixes #488
Added display of timestamps to odometry data and a config variable, "Timestamp Interval", to configure it.

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/odometry_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/odometry_plugin.h
@@ -67,11 +67,17 @@ namespace mapviz_plugins
     {
     }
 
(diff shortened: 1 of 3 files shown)


jatowler:
This PR does two things:

1. Fixes #488 
2. Adds timestamp display to the odometry plugin

Please split this into two different PRs: one that fixes the issue, and another that adds the feature. Also, please file an issue describing the new feature, and then reference it in the second PR.

In general, you can tell a PR should be split when the title has the word "and" in it.

Issue #494, opened 2017-07-20
Add a way to view the covariances at each odometry point


matt-attack:
Current UI only allows for viewing the most recent covariance. Would be useful if there was a way to view the covariance at each of the recorded points to be able to easily see how it changes over time and space.

PR #495, opened 2017-07-20
Add a way to show all odometry covariances

matt-attack:
Current UI only allows for viewing the most recent covariance. Added a
checkbox and functionality to enable viewing the covariance at each
point to help with identifying how it changes over time.
Implements #494

Depends on Merge Request #493


mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
       PrintInfo("OK");
     }
   }
+  
+
+  void OdometryPlugin::Paint(QPainter* painter, double x, double y, double scale)
+  {
+    int interval = ui_.show_timestamps->value();
+    if (interval == 0)
```
jatowler: "Naked if" statements are dangerous for a number of reasons, primary among which is that a future modification may fail to notice the lack of braces and attempt to add a statement to this block. In this case, either that statement will only execute when the condition is false, which is the exact opposite of what is intended, or the function will *always* return, which is not desired either.

Always use braces to denote blocks, even if they are only one line.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
...
+
+    //set the draw color for the text to be the same as the rest
+    QPen pen(QBrush(ui_.color->color()), 1);
+    painter->setPen(pen);
+
+    std::list<StampedPoint>::iterator it = points_.begin();
+    int i = 0;//used to alternate between rendering text on some points
+    for (; it != points_.end(); ++it)
+    {
+      if (it->transformed && i++ % interval == 0)
```
jatowler: Side-effect increment/decrement operators are dangerous for a number of reasons; in this case, because their behavior is too easy to parse incorrectly. Presumably, you intend the postfix behavior here, but there is no good reason to be unclear. In general, prefer code clarity over brevity.

Move the increment outside of the conditional.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
...
+      {
+        QPointF point = tf.map(QPointF(it->transformed_point.getX(),
+                                       it->transformed_point.getY()));
+		QString time;
```
jatowler: It appears your editor has accidentally added some tab characters here. Files in `swri-robotics` use spaces instead of tabs, so it is good to match the existing style.

Replace tabs with two spaces each.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
...
-    {
-      glBegin(GL_LINE_STRIP);
+    glColor4d(color_.redF(), color_.greenF(), color_.blueF(), 1.0);
 
-      for (uint32_t i = 0; i < cur_point_.transformed_cov_points.size(); i++)
+    std::list<StampedPoint>::iterator it = points_.begin();
+    for (; it != points_.end(); ++it)
+    {
+      auto cur_point = *it;
+      if (ui_.show_all_covariance->isChecked() == false)
```
jatowler: See elsewhere about the naked if.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
...
 
-      glVertex2d(cur_point_.transformed_cov_points.front().getX(),
-                 cur_point_.transformed_cov_points.front().getY());
+        glVertex2d(cur_point.transformed_cov_points.front().getX(),
+                   cur_point.transformed_cov_points.front().getY());
 
-      glEnd();
+        glEnd();
+      }
+      if (ui_.show_all_covariance->isChecked() == false)
```
jatowler: See elsewhere about the naked if.

jatowler: Kudos for the explicit check against `false`.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -411,5 +467,10 @@ namespace mapviz_plugins
     emitter << YAML::Key << "static_arrow_sizes" << YAML::Value << ui_.static_arrow_sizes->isChecked();
 
     emitter << YAML::Key << "arrow_size" << YAML::Value << ui_.arrow_size->value();
+
+    emitter << YAML::Key << "show_timestamps" << YAML::Value << ui_.show_timestamps->value();
```
jatowler: This YAML key doesn't appear to have anything to do with showing all covariance ovals.

Please constrain changes in a PR to a single issue.

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -208,6 +193,21 @@ namespace mapviz_plugins
         }
       }
     }
+
+    if (points_.empty() ||
+        stamped_point.point.distance(points_.back().point) >=
```
jatowler: In complex conditionals like this, prefer to explicitly using parentheses instead of relying on order of operations. Other coders may not necessarily know the order of operations, and omitting the parentheses forces readers to parse this entire expression before understanding its meaning. Optimize code for reading.

Please surround the `>=` expression with parentheses.

mapviz_plugins/src/odometry_plugin.cpp
jatowler: It would be good to explain in a comment *why* you return when `interval == 0`. I suspect it is because we do not wish to paint all the covariance ovals when the interval is zero, but a) a future reader would appreciate that comment; and b) why is 0 the magic number? (For example, I could imagine a reader thinking that an interval of 0 meant to draw every oval, and an interval of 1 meant to draw every other; interval of 2 to draw every third, etc.; this coder would think -1 is the correct magic number for don't-draw-any-ovals.)

jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -278,25 +278,70 @@ namespace mapviz_plugins
...
+    std::list<StampedPoint>::iterator it = points_.begin();
+    int i = 0;//used to alternate between rendering text on some points
+    for (; it != points_.end(); ++it)
+    {
+      if (it->transformed && i++ % interval == 0)
+      {
+        QPointF point = tf.map(QPointF(it->transformed_point.getX(),
+                                       it->transformed_point.getY()));
+		QString time;
+		time.setNum(it->stamp.toSec(), 'g', 12);
```
jatowler: See above about tab characters

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
```diff
@@ -377,6 +422,17 @@ namespace mapviz_plugins
       ui_.arrow_size->setValue(node["arrow_size"].as<int>());
     }
 
+    if (node["show_timestamps"])
```
jatowler: This does not appear to have anything to do with showing covariances.

Please constrain changes in a single PR to a single issue.

mapviz_plugins/src/odometry_plugin.cpp
jatowler: Adding to review

mapviz_plugins/src/odometry_plugin.cpp
jatowler: It occurs to me that this PR may depend on #493, in which case that should be explicitly stated in the PR text so we can review the PRs in order.

matt-attack:
Yep!
image

PR #496, opened 2017-08-02
Add missing boost header.

malban:
Add missing boost header.

It looks like a change in marti_common exposed that we need to include boost/algorithm/string/trim.hpp


evenator:
Related: #497 (which makes the same fix in kinetic)

PR #504, opened 2017-08-14
Support transparent tiles in multires_image (Kinetic)

### LINKED ISSUE #501: [multires_image] Transparent Regions Rendered in Black
Hello,
When I try to overlay an image on a tile map in mapviz, I notice the transparent regions of the image get colored black(I am using mapproxy with google map layer). If I put the same image on the google-earth application (with kml file), the transparent region is displayed correctly(I see the google map layer). Are there any ideas on how I can resolve this?

pjreed:
The multires_image plugin was internally loading textures without an
alpha channel, so tiles that had transparent regions would show
black instead.  This will cause transparent regions to render properly.

Fixes #501

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ msg_gen/
 srv_gen/
 *.cfgc
 version.txt
+cmake-build-debug/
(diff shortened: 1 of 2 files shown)


evenator:
This looks good to me.

PR #506, opened 2017-09-06
Remove dep on libopencv-dev from multires_image (Kinetic/Lunar)

pjreed:
It doesn't actually use it at all, and it's the only thing preventing the
kinetic-devel branch from compiling on lunar.

### DIFF
--- a/multires_image/package.xml
+++ b/multires_image/package.xml
@@ -13,11 +13,10 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>libopencv-dev</depend>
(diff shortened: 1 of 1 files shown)


multires_image/package.xml
```diff
@@ -13,11 +13,10 @@
 
   <buildtool_depend>catkin</buildtool_depend>
 
-  <depend>libopencv-dev</depend>
+  <build_depend>libqt5-opengl-dev</build_depend>
```
evenator: I find the package dependencies easier to parse when sorted by package, rather than dependency type.

PR #512, opened 2017-09-27
Added timestamp display to odometry for kinetic

matt-attack:


pjreed:
Looks like this one is missing commit 76ee092.

PR #514, opened 2017-10-25
change the signal that triggers AlphaEdited + minor changes

facontidavide:
Hi, 

I consider counter intuitive the fact that you must press ENTER to update Alpha value in the UI.
For this reason I suggest using the signal valueChanged(double).

Furthermore:

- I limit the size of buffer in the range [1,100] and 
- use the method value() in QDoubleSpinBox instead of text()->toDouble().


facontidavide:
Please note that the fact this also solves a BUG that trigger Aplha to become 0.0 when we launch the application.

pjreed:
Tested it and this works well for me.  Thanks!

PR #515, opened 2017-10-26
change the signal that triggers AlphaEdited + minor changes (#514)

pjreed:
* pointcloud and laserscan plugins: change the signal that triggers AlphaEdited + minor changes
* pointcloud and laserscan plugins: need_minmax_ was not updated correctly

### DIFF
--- a/mapviz_plugins/ui/pointcloud2_config.ui
+++ b/mapviz_plugins/ui/pointcloud2_config.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>306</width>
+    <width>307</width>
     <height>341</height>
    </rect>
   </property>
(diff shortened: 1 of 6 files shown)


mapviz_plugins/ui/pointcloud2_config.ui
```diff
@@ -102,8 +111,11 @@
    </item>
    <item row="3" column="1">
     <widget class="QSpinBox" name="bufferSize">
+     <property name="minimum">
+      <number>1</number>
```
evenator: Isn't a minimum buffer size used for an infinite buffer?

malban: That was the intent for odometry, gps, tf, etc. data, but I'm not sure if an infinite buffer makes much sense for point cloud data.  It doesn't take too long for point cloud data to accumulate to the point it becomes impractical to render.

pjreed: In practice, the pointcloud2 plugin is actually really slow.  Having a buffer of more than 10 already has serious performance issues for large point clouds; allowing an infinite buffer was a good way to make Mapviz quickly peg an entire CPU core and then grind to a halt.

PR #519, opened 2017-10-31
Re-add GPSFix plugin to kinetic-devel

pjreed:
The GPSFix plugin had been removed from the kinetic-devel branch long
ago because the gps_common package hadn't been released into Kinetic
at the time; it exists now, so there's no reason for this plugin to
not exist.


evenator:
LGTM, but I haven't tested it yet.

PR #521, opened 2017-10-31
Add support for unpacking rgb8 in pointcloud2s (Kinetic)

matt-attack:


pjreed:
Out of curiosity, exactly what does this do?

matt-attack:
Some pointclouds contain a per point RGB8 color that is encoded into a float. This adds an option to color the points based on this decoded color. 
Somehow RVIZ handles this but I found out that mapviz does not.

PR #522, opened 2017-11-01
Add keyboard input support for plugins (Indigo)

matt-attack:


pjreed:
Before I merge this in, note that since this changes the MapCanvas API, any plugins that directly access the MapCanvas object will need to be recompiled or else they will probably crash on load.  (This might catch you by surprise if you're compiling Mapviz from source but have plugins from another package that are binary-installed)

@malban  @evenator @kriskozak

PR #525, opened 2017-11-06
Speed up improvement in LaserScan and PointCloud2

facontidavide:
Hi,

it is possible to reduce dramatically the memory allocation in both **std::vector::reserve** or recycling instances popped from the **scan_** buffers.

In my benchmarks, these change roughly reduce the amount of CPU used by these two plugins to 1/3rd.

### DIFF
--- a/mapviz_plugins/src/laserscan_plugin.cpp
+++ b/mapviz_plugins/src/laserscan_plugin.cpp
@@ -324,6 +324,31 @@ namespace mapviz_plugins
     point_size_ = static_cast<size_t>(value);
   }
 
+  void LaserScanPlugin::updatePreComputedTriginometic(const sensor_msgs::LaserScanConstPtr& msg)
+  {
+      static size_t prev_size      = 0;
+      static float  prev_angle_min = msg->angle_min;
+      static float  prev_increment = msg->angle_increment;
+
+      if( msg->ranges.size() != prev_size ||
+          msg->angle_min !=  prev_angle_min  ||
+          msg->angle_increment != prev_increment   )
+      {
+          prev_size = msg->ranges.size();
+          prev_angle_min = msg->angle_min;
+          prev_increment = msg->angle_increment;
+          precomputed_cos_.resize( prev_size );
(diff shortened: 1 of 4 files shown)


mapviz_plugins/src/laserscan_plugin.cpp
```diff
@@ -324,6 +324,31 @@ namespace mapviz_plugins
     point_size_ = static_cast<size_t>(value);
   }
 
+  void LaserScanPlugin::updatePreComputedTriginometic(const sensor_msgs::LaserScanConstPtr& msg)
+  {
+      static size_t prev_size      = 0;
+      static float  prev_angle_min = msg->angle_min;
```
evenator: Please avoid the use of local static variables by refactoring these to class variables or creating a new object to encapsulate the entire trigonometry cache.

facontidavide: I understand this is "personal taste". Or is there any technical reason to avoid static?
No problem, though

facontidavide:
I applied the suggested changes, even if philosophically I disagree ;)
"Your house, your rules".

Cheers and thanks again for the amazing tool.

mapviz_plugins/src/laserscan_plugin.cpp
facontidavide: I understand you point now. You are right

facontidavide:
For the records:

1) using __std::vector::reserve__ or __std::vector::resize__ is always good.
2) In laserscan, the main performance killer was mostly the calculation of __cos()__ and __sin()__. 
3) the object __LaserScanPlugin::Scan__ is relatively fast to copy in comparison to __PointCloud2Plugin::Scan__, which is pretty slow.
4) The worst in terms of performance in __PointCloud2Plugin__ was that ANY single point required a memory allocation because of  __PointCloud2Plugin::StampedPoint::features__. Recycling an already allocated instance avoid that.

pjreed:
This looks good to me.  Tested and works in my Kinetic Docker environment.  Thanks!

PR #526, opened 2017-11-06
add context menu to config_item

facontidavide:
Hi,

I found the Remove button very counter intuitive, in particular when the config_item is collapsed.
I hope you will find this PR useful.
image

Regards

Davide


evenator:
👍  From me. I've always found that button unintuitive myself.

facontidavide:
Just a personal opinion. If this PR is eventually merged, may be you want to consider removing the __Remove__ button on bottom and stop using orange to highlight the selected config_items.

Up to you ;)

pjreed:
:+1: 

I like this, although I don't think the "Remove" button on bottom should be removed.  In general I think that a context menu should never be the only way to access a function, since right-clicking not be available or may not work right on every platform, and it's not easily discoverable for users who don't expect to be able to right-click on everything.

PR #527, opened 2017-11-07
Improvement of image plugin. Transport is selected automatically for compressed images

facontidavide:
Hi,

as a newbie a spent quite a lot of time understanding how to use correctly the image plugin when my image is compressed.

In my case for, instance,I have a __/zed/right/image_raw/compressed__.

 I have to first select transport "compressed" in the combo box and then type manually __/zed/right/image_raw__ because the topic selector doesn't work.

With this PR the topics of types __sensor_msgs/CompressedImage__ is visible in the topic selector.
If one of those topics is selected, the suffic of the topic name is used to automatically select the proper combo box ("compressed", "theora", "compressedDepth").

Cheers

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/image_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/image_plugin.h
@@ -112,7 +112,7 @@ namespace mapviz_plugins
     void SetWidth(int width);
     void SetHeight(int height);
     void SetSubscription(bool visible);
-    void SetTransport(const QString& transport);
+    void SetTransport(int index);
 
   private:
     Ui::image_config ui_;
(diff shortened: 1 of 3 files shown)


facontidavide:
By the way, I also found some minor issues when the visibility of the config_item change and the image is compression (resubscription didn't work).

facontidavide:
Hi,

I see your point, especially for __theora_image_transport__, but I still believe that there must be a better way to simplify usage.
For sure the PR I am proposing need improvements.

Nevertheless, is disagree with this sentence:

> f you have a publisher that is only publishing a CompressedImage and not using the image_transport mechanism, I think it would be preferable to fix that so that it also publishes the raw topic

I might have my reasons to avoid that (WIFI connection).

Apparently, ImagePlugin can use the local_node_ to deal with this automatically and locally.
https://github.com/swri-robotics/mapviz/blob/kinetic-devel/mapviz_plugins/src/image_plugin.cpp#L263

What bother me most is that the Topic Selector is useless in these cases, because it will remain empty.
I will keep thinking about it...

facontidavide:
I will close this for the time being...

PR #533, opened 2017-11-09
missing "arrows" in NavSat plugin

facontidavide:
Hi,

sorry if I bother you again, but I guess this was either a bug or dead code.
I am using NavSat plugin and despite the fact that I see options to resize the arrow (?) there is no "arrows" option in the combo box.

Cheers

Davide

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/navsat_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/navsat_plugin.h
@@ -72,6 +72,7 @@ namespace mapviz_plugins
    protected Q_SLOTS:
     void SelectTopic();
     void TopicEdited();
+    virtual void SetDrawStyle(QString style) override;
 
    private:
     Ui::navsat_config ui_;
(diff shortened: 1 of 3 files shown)


facontidavide:
By the way, if for any reason it was __intentional__ to remove the "ARROW" visualization from this plugin, I suggest to temporary hide these widgets 

      ui_.label_arrow->setVisible(false);
      ui_.arrow_size->setVisible(false);
      ui_.static_arrow_sizes->setVisible(false);

pjreed:
It was actually intentional to remove it; the [sensor_msgs/NavSatFix](http://docs.ros.org/api/sensor_msgs/html/msg/NavSatFix.html) message does not have any heading information in it, so drawing an arrow for it would be misleading because it doesn't actually have a direction.  References to it are probably there just because it, the GPSFix, and the Odometry plugins are all very similar and have had a lot of code copied and pasted between them.  I think it'd make sense to hide the arrow-related widgets.

facontidavide:
This is what I guessed. Shall we remove the dead code then ;)?

PR #539, opened 2017-11-14
Glew warning fixed

facontidavide:
Hi,

this PR should reduce the number of warnings related to glew.h

Cheers

PR #543, opened 2017-11-16
Added "keep image ratio" to Image plugin

facontidavide:
Hi,

I believe this PR would help making the Image Plugin more usable.

1) Added "Keep original image ratio" to the Widget. When checked, the height is automatically adjust, both in "percent" and "pixel" mode. The user should only modify the __Width__ whilst the Height is updated accordingly.

2) When changing the mode from "percent" to "pixel" and vice versa, the image size is preserved.

3) To be sure that no resize is introduced in step 2), I had to change the width and height __type__ and QSpinBox to double and QDoubleSpinBox, and percend must have 1 decimal.

Cheers

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/image_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/image_plugin.h
@@ -109,10 +109,11 @@ namespace mapviz_plugins
     void SetUnits(QString units);
     void SetOffsetX(int offset);
     void SetOffsetY(int offset);
-    void SetWidth(int width);
-    void SetHeight(int height);
+    void SetWidth(double width);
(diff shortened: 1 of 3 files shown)


facontidavide:
https://vimeo.com/243268373

facontidavide:
Does anyone at SWRI ever say "thanks for contributing"?

malban:
Some of us do at least: https://github.com/swri-robotics/mapviz/pull/539#issuecomment-345371749 https://github.com/swri-robotics/mapviz/pull/525#issuecomment-342520534

But regardless, thanks for contributing, we appreciate it and are glad you find the tool useful.

Issue #529, opened 2017-11-08
LaserScan transform


facontidavide:
Hi,

using LaserScan I noticed that the transform operation is potentially applied twice, once in the __laserScanCallback__ and later, before, drawing, in __Transform()__.

I am not sure I understand the reasoning behind repeating the transform in two different places.
In my case, the __GetTransform__ succeed in __laserScanCallback__ and fails in __Tranform()__.
Therefore Draw() will not display it. https://github.com/swri-robotics/mapviz/blob/indigo-devel/mapviz_plugins/src/laserscan_plugin.cpp#L452

shouldn't Transform() be modified as follows:

``` c++
  void LaserScanPlugin::Transform()
  {
    std::deque<Scan>::iterator scan_it = scans_.begin();
    for (; scan_it != scans_.end(); ++scan_it)
    {
      Scan& scan = *scan_it;

      swri_transform_util::Transform transform;

      if( !scan.transformed )
      {
        bool was_using_latest_transforms = this->use_latest_transforms_;
        this->use_latest_transforms_ = false;
        if (GetTransform(scan.source_frame_, scan.stamp, transform))
        {
          scan.transformed = true;
          std::vector<StampedPoint>::iterator point_it = scan.points.begin();
          for (; point_it != scan.points.end(); ++point_it)
          {
            point_it->transformed_point = transform * point_it->point;
          }
        }
        this->use_latest_transforms_ = was_using_latest_transforms;    
      }
    }
    // Z color is based on transformed color, so it is dependent on the
    // transform
    if (ui_.color_transformer->currentIndex() == COLOR_Z)
    {
      UpdateColors();
    }
  }
``` 

Cheers

facontidavide:
any feedback?

malban:
That makes sense to me.  It looks like there is an inconsistency in the two code paths though.  In the callback the use_latest_transforms flag is left whatever it is, but in the Transform function it is set to false explicitly.  That may be why you are seeing it fail in one and not the other.

I think in almost every conceivable case the use_latest_transforms flag should be false for this type of data.  The use_latest_transforms flag ignores the stamp of the data resulting in the points being transformed to the current sensor position instead of the sensor position when the data was captured.  So, in addition to your suggested change, we would also want to make sure the correct transform is applied in the callback.

facontidavide:
Al right. my problem is that __unless__ I use latest transform, my laser fails.

[...]

PR #544, opened 2017-11-17
Dont transform twice laser

facontidavide:
Pull request related to issue #529.


facontidavide:
I realized that I had to take into account the signal TargetFrameChanged...
Fixed.

Issue #549, opened 2017-11-30
Initial position and 2D goal buttons


facontidavide:
Hi,

For my application, it would be very convenient to add these buttons that are available in RViz. 

![image](https://user-images.githubusercontent.com/2822888/33426229-66236b6c-d5c1-11e7-987a-7a87f7378fa7.png)

I can work on this and send a PR. My question is where do you think that they be added.

1) A separate plugin, not related to core application?
2) __statusbar__ on the bottom?
3) Actions in the __menubar__ on top?
4) __widget_2__ inside __configdock__?

Cheers

malban:
It should definitely be a plug-in and not part of the core application.  For now, the best place for it would be like any of the other plug-ins.  In the future we may expand on the current plug-in architecture to provide access to more GUI elements like the menubar and statusbar.

PR #551, opened 2017-12-01
New plugin to send commands to move_base

facontidavide:
Hi,

this first version of the plugin is a little rough around the edges but it does its job, IMO. #549 

https://vimeo.com/245356427

It is intentionally modeled after the arrows on top of RViz

Cheers

Davide


mapviz_plugins/include/mapviz_plugins/move_base_plugin.h
```diff
@@ -0,0 +1,124 @@
+// *****************************************************************************
+//
+// Copyright (c) 2014, Southwest Research Institute® (SwRI®)
```
pjreed: Update the year to 2017 here, since that's when this was created.

mapviz_plugins/include/mapviz_plugins/move_base_plugin.h
```diff
@@ -0,0 +1,124 @@
...
+
+// Mapviz libraries
+#include <mapviz/map_canvas.h>
+
+#include <geometry_msgs/PoseWithCovarianceStamped.h>
+#include <move_base_msgs/MoveBaseAction.h>
+#include <actionlib/client/simple_action_client.h>
+
+// QT autogenerated files
+#include "../mapviz_plugins/ui_move_base_config.h"
```
pjreed: This should just be `#include "ui_move_base_config.h"`; CMake will include the auto-generated header path automatically.

mapviz_plugins/include/mapviz_plugins/move_base_plugin.h
```diff
@@ -0,0 +1,124 @@
...
+    QWidget* config_widget_;
+    mapviz::MapCanvas* map_canvas_;
+
+    ros::NodeHandle nh_;
+    ros::Publisher init_pose_pub_;
+
+    bool is_mouse_down_;
+    QPointF arrow_tail_position_;
+    float arrow_angle_;
+    qint64 mouse_down_time_;
```
pjreed: `mouse_down_time_` is unused; does it need to be here?

mapviz_plugins/src/move_base_plugin.cpp
```diff
@@ -0,0 +1,338 @@
...
+        init_pose_pub_.publish(initpose);
+        ui_.pushButtonInitialPose->setChecked(false);
+    }
+    return true;
+}
+
+
+void MoveBasePlugin::Draw(double x, double y, double scale)
+{
+    const QPointF arrow_points[7] = {
```
pjreed: I think it'd be a good idea to pull this 7 out into a `constexpr` since it is used multiple times here and it wouldn't make sense to change it once without changing it elsewhere.

mapviz_plugins/src/move_base_plugin.cpp
```diff
@@ -0,0 +1,338 @@
...
+        {
+            tf::Vector3 point(arrow_points[i].x(), arrow_points[i].y(), 0);
+            point *= scale*10;
+            point = tf::Transform( tf::createQuaternionFromYaw(arrow_angle_)  ) * point;
+            transformed_points[i] = QPointF(point.x() + arrow_tail_position_.x(),
+                                            point.y() + arrow_tail_position_.y() );
+        }
+        glColor3f(0.1, 0.9, 0.1);
+        glBegin(GL_TRIANGLE_FAN);
+        for (size_t i = 0; i<7; i++)
```
pjreed: It'd be preferable to use a range-based `for` here (like `for (const auto& point : transformed_points) { ... }`).

mapviz_plugins/src/move_base_plugin.cpp
```diff
@@ -0,0 +1,338 @@
+#include <mapviz_plugins/move_base_plugin.h>
```
pjreed: Please also include the BSD license header at the beginning of this file.

mapviz_plugins/src/move_base_plugin.cpp
```diff
@@ -0,0 +1,338 @@
...
+        move_base_msg_.action_goal.goal_id.id = "mapviz_goal";
+        move_base_msg_.action_goal.goal.target_pose.header = move_base_msg_.action_goal.header;
+
+        geometry_msgs::Pose& pose = move_base_msg_.action_goal.goal.target_pose.pose;
+        pose.position.x = arrow_tail_position_.x();
+        pose.position.y = arrow_tail_position_.y();
+        pose.position.z = 0.0;
+        tf::quaternionTFToMsg( quat, pose.orientation );
+
+        move_base_client_.sendGoal(move_base_msg_.action_goal.goal);
```
pjreed: I see that you're calling a `move_base` service here, but it looks like the rviz "2D Nav Goal" button publishes a `geometry_msgs/PoseStamped` message.  Why the difference?

mapviz_plugins/src/move_base_plugin.cpp
facontidavide: you read my mind. I love range based for, but I noticed that the rest of the code don't use it much.

pjreed: Many of the plugins were written before Mapviz supported C++11, so they still use old-style for loops, but I prefer range-based loops for new code.

mapviz_plugins/src/move_base_plugin.cpp
facontidavide: the RViz interface is "send and forget". It means that I can't get any feedback about the current execution (getStatus) and I won't be able to cancel the action.

pjreed:
Also, please update the README.md with some information about the plugin.

facontidavide:
I applied the changes you suggested. THANKS!

pjreed:
Looks good to me.  Thanks again!

Issue #542, opened 2017-11-16
Suggestion: Image plugin "percent" should mean percent of original size


facontidavide:
Hi,

I found the usability of the image "percent" option a little bit confusing.
In general, I expect the majority of the user to want the image ratio to be preserved.

I was preparing a PR to add a "Keep original image ratio" checkbox to the GUI, but the code is very hard to reason about because of the fact that "percent" is percent of the canvas.

Would you consider to change this to "percent of the original image size" ?

facontidavide:
I guess it is a "no"

malban:
I think the change is reasonable and would welcome a pull request.  It is unlikely that we will address the issue ourselves though.

PR #553, opened 2017-12-01
Image plugin: "percent" should refer to original image size

facontidavide:
Change proposed in #542 

I also changed few attributes to types QPointF and QSizeF, that, I know, might be an opinionated decision.


malban:
I like this, but can we keep the original behavior as an option.  It is sometimes useful to have the image scale with the window.  The options could be: Pixel, Percent, Percent of Canvas or something.

facontidavide:
It make sense, but honestly I am prioritizing other improvements that are more relevant for may project, therefore I will not be able to focus on enhancing this PR as you suggest until the second half of January.

malban:
No worries.

facontidavide:
I will temporarily close this PR, to focus better on the others PR still ones

PR #555, opened 2017-12-17
Sharing tf_manager_ between main app and plugins

facontidavide:
Hi,

this is a non-functional change. 
In short it share the instance of **tf_manage_** with the plugins during initialization.
It also removed the local instances of **local_xy_util_** in the gps, navsat tile_map plugins.


facontidavide:
note. This branch seems to fail because it relies on the PR of swri_transform_utils

https://github.com/swri-robotics/marti_common/pull/501

facontidavide:
with the recent sync of the kinetic and indigo repository, this shouldn't fail any more to compile

pjreed:
Cool, this seems to work really well and improves startup time by quite a bit when you are loading lots of plugins.  It breaks compatibility with the current plugin API, though; since I have recently merged in a lot of changes, I will probably make a release of the current version and then wait until the after the next monthly ROS update to make a release with this.

facontidavide:
I didn't see this coming...  after one year!

Just kidding, I am happy you found it useful. Cheers

PR #556, opened 2017-12-27
Bug fix in TF plugin. Can't calculate distance if buffer is empty

facontidavide:
Hi,

the TF plugin used to work, but today it crushed with no apparent reason.
I found a bug: distance can not be calculated unless **points_.empty() == false**


facontidavide:
I will close this, since I can improve even more #557

PR #557, opened 2017-12-27
Improving point_drawing plugins and bug fix of tf_plugin

facontidavide:
- Don't transform more than once a StampedPoint.
- use std:.deque instead of std::list for performance improvements


facontidavide:
my bad, this change break the buffered lines when the fixed frame is changed. Let me apply a couple of improvements. 

Cheers

facontidavide:
Fixed

facontidavide:
do not merge yet

facontidavide:
Alright. With the last commit this PR became slightly more opinionated than it was intended to be.

I tried to move some complexity from the derived classes of PointDrawingPlugin to the latter.
To do so, I changed the permission of some attributes from protected to private and increased the number of protected methods and slots.

Arguably the most significant improvement is the addition of **pushPoint** to reduce copy and pasting.

This prevents a potential crash in **tf_plugin**, as reported in #556

Let me know what you think :)

mapviz_plugins/src/point_drawing_plugin.cpp
```diff
@@ -464,11 +563,11 @@ namespace mapviz_plugins
       for (size_t i = 0; i < laps_.size(); i++)
       {
         UpdateColor(base_color, static_cast<int>(i));
-        std::list<StampedPoint>::iterator it = laps_[i].begin();
-        for (; it != laps_[i].end(); ++it)
+        std::deque<StampedPoint>::iterator it = laps_[i].begin();
```
malban: Is this line necessary now?

facontidavide:
No, I forgot to delete it

facontidavide:
line deleted

matt-attack:
When testing this on my system it seems that changing the draw style of point drawing mods no longer functions, it always draws lines. Does it work properly on your end?

facontidavide:
I will check it out

facontidavide:
Fixed. Sorry about that ;)

facontidavide:
I am adding yet another commit to fix a broken SIGNAL/SLOT pair in odometry_plugin

matt-attack:
Just tested it and now changing the draw style works but arrows do not rescale when you zoom in or out.

facontidavide:
good catch.

Apparently the problem is related to the fact that I am trying to avoid recomputing the transform more than necessary. 
https://github.com/Eurecat/mapviz/blob/point_plugin/mapviz_plugins/src/point_drawing_plugin.cpp#L384

Let me find a solution :)

facontidavide:
Ok, it should be fixed now. 
This branch now uses 35% less CPU than the original one for a buffer length or 100/200 points.

facontidavide:
"Points in any style also don't seem to render in their correct location either."

Is this still true with the last commit? To me the points look just right

facontidavide:
If these problems persist and you think that the PR is becoming too complex to merge, let me know.

At the end of the day the original goal was only to avoid calling transform too often and to prevent crashes in tf_plugin :(

malban:
Improving the speed and fixing crashes seems like a worthy goal :)

facontidavide:
I just pushed an additional commit that fixes some issues and adds a button to clear the buffer.

PR #558, opened 2017-12-28
OpenGL rendering of PointClouds  (2X speedup)

facontidavide:
Hi,

working on another plugin (that I will submit soon) I figured out that we may use a more modern OpenGl API as described here:

https://cognitivewaves.wordpress.com/opengl-vbo-shader-vao/

I decided to try it on a plugin that I know well that display a lot of points, i.e. PointCloud2.
On my machine, under heavy load (Velodyne points with buffer size 10), the CPU usage is halved.

I also took the opportunity to use C++11 range-loops here and there.

Cheers

Davide


facontidavide:
I took the opportunity to add some improvements.

1) Use less memory, using bytes instead of floats to represent colors.
2) Added a button to clear the buffer (other PR of mines will also include that in the future)
3) Unsubscribe from topic when the widget is marked as "not visible" 

The latter one might be controversial, but I have a good argument ;)
PointCloud2 might be a very large message and I believe that many users will connect to the robot using Mapviz remotely, using Wifi connection.

When the plugin is disabled, it is fair to expect that no bandwidth is used.

let me know what you think :)

pjreed:
Thanks, this looks good.  I also agree that it's a good idea to unsubscribe from a topic when a plugin isn't visible; I guess that was just missed on this plugin.

PR #559, opened 2018-01-03
Bug fix in TileMap. GenTexture was invoked over and over again

facontidavide:
TileMap used sometime 100% of the CPU.
After some profiling I noticed that GenTexture was called many times, more than I expected.

Inspecting the code I found this bug. Now the amount of CPU is considerably lower and related mostly to rendering, not to GenTexture.

### DIFF
--- a/tile_map/src/tile_map_view.cpp
+++ b/tile_map/src/tile_map_view.cpp
@@ -203,6 +203,59 @@ namespace tile_map
     }
   }
 
+  void TileMapView::DrawTiles(std::vector<Tile>& tiles, int priority)
+  {
+    for (size_t i = 0; i < tiles.size(); i++)
+    {
+      TexturePtr& texture = tiles[i].texture;
+
+      if (!texture)
+      {
+        bool failed;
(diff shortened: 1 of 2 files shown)


tile_map/src/tile_map_view.cpp
```diff
@@ -214,7 +214,7 @@ namespace tile_map
 
     for (size_t i = 0; i < precache_.size(); i++)
     {
-      TexturePtr texture = precache_[i].texture;
+      TexturePtr& texture = precache_[i].texture;
```
malban: It isn't clear to me how this would change the behavior of the code here.

The original code is copying the shared ptr, but that is relatively cheap and shouldn't invoke a genTexture call unless the shared ptr is null.  The new code avoids the ptr copy but shouldn't reduce genTexture calls.

Is there something else going on here?

malban: Ah, I see the issue.  Without the reference the new texture object created isn't saved back to the cached objects.

malban:
Tested these changes on Indigo.

malban:
Thanks @facontidavide

PR #561, opened 2018-01-03
New plugin: OccupancyGrid / Maps

facontidavide:
Hi,

I developed this for my own project, but this might be useful to other people.

Cheers

Davide

image


facontidavide:
not ready for prime-time. It uses too much RAM or/and CPU. I have a solution in mind though

malban:
@facontidavide You should be able to render the occupancy grid as a textured quad, much like the textured marker plugin and should get better performance.

PR #563, opened 2018-01-08
Bug fix in image plugin

facontidavide:
Hi,

using the Image Plugin with a robot connected on a Wifi network, I noticed that much more bandwidth than expected was frequently used.

I found that when the visibility changes (i.e., the signal VisibleChanged is emitted) the slot SetSubscription is executed. Unfortunately this method always subscribe without transport hint, therefore the raw format is __always__ used instead of a compressed one, when compression is selected by the user.

The first commit fixes this issue, whilst the second just apply some code simplification that doesn't have any real effect.


pjreed:
This looks good, thanks for adding it.  Sorry it's taken so long to get to this, I haven't had time to work on mapviz for a while now.

PR #565, opened 2018-01-18
Add ability to set offset for multires image

### LINKED ISSUE #564: Add ability to shift multires image by an x,y offset
Often, map data from image.geo files can be off by a couple meters. I think it would be helpful to have the ability to shift the entire map by some x,y offset, so a user can visually align the map without having to modify the underlying data.

jgassaway:
Closes #564 by adding ability to set an x,y offset (applied in Easting and Northing) to a multires image to correct for inaccuracy in the original map data.

Example of map data that is inaccurate:
mapviz-add-multires-before

Corrected map data using offset:
mapviz-add-multires-after

### DIFF
--- a/multires_image/include/multires_image/multires_image_plugin.h
+++ b/multires_image/include/multires_image/multires_image_plugin.h
@@ -81,11 +81,15 @@ namespace mapviz_plugins
   protected Q_SLOTS:
     void SelectFile();
     void AcceptConfiguration();
+    void SetXOffset(double long_offset);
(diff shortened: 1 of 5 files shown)


malban:
Looks good.  Can you also make a pull request for Jade?

pjreed:
Side note: Jade isn't supported any more and there won't be any more releases for it.  Feel free to keep merging things into `jade-devel`, but only `indigo-devel` and `kinetic-devel` actually matter going forward.

Issue #566, opened 2018-01-22
multires_image only shows one layer


kimmat:
Hi,

I have generated a multires_image with six different layers, but mapviz only visualize the first layer. I am using Ubuntu 16.04 with the following packages:

```
ros-kinetic-mapviz           0.2.4-0xenial-20180105-174837-0800
ros-kinetic-mapviz-plugins   0.2.4-0xenial-20180105-194002-0800
ros-kinetic-multires-image   0.2.4-0xenial-20180105-180349-0800
```

Here is my config file

```
image_path: "."
image_width: 8192
image_height: 8192
tile_size: 256
datum: wgs84
projection: wgs84
tiepoints:             #   [pixel x, pixel y, geo x, geo y]
 - point: [0, 0, 11.118164, 60.261617]
 - point: [8192, 0, 11.162109, 60.261617]
 - point: [0, 8192, 11.118164, 60.239811]
 - point: [8192, 8192, 11.162109, 60.239811]
```

Any idea of what might cause this error? is it a bug, or am I doing something wrong?

kimmat:
Found the error. Looks like the min_scale_ parameter was calculated using lat/long coordinates, but it should be calculated with local coordinates in meters.

Here are the changes I made in ```multires_image/src/multires_view.cpp``` in the constructor of ```MultiresView```

```cpp
  MultiresView::MultiresView(multires_image::TileSet* tiles, QGLWidget* widget) :
      m_tiles(tiles),
      m_cache(tiles, widget),
      m_currentLayer(tiles->LayerCount() - 1),
      m_startRow(0),
      m_startColumn(0),
      m_endRow(0),
      m_endColumn(0)
  {
    double top, left, bottom, right;
    double local_width, local_height;
    tiles->GeoReference().GetCoordinate(0, 0, left, top);

    tiles->GeoReference().GetCoordinate(
      tiles->GeoReference().Width(),
      tiles->GeoReference().Height(),
      right,
      bottom);

    swri_transform_util::LocalXyFromWgs84(top, left, bottom, right, local_width, local_height);

    double scale_x = local_width / tiles->GeoReference().Width();
    double scale_y = local_height / tiles->GeoReference().Height();

    min_scale_ = scale_x;
    if (scale_y > scale_x)
      min_scale_ = scale_y;
  }
```

I would be very appreciative if you could include this patch into the main repository.

PR #569, opened 2018-02-02
Fix multires image scale when projection is WGS84.

malban:
Refs #566


malban:
@kimmat Can you confirm this branch resolves the issue for you?

kimmat:
I can confirm that this solved my issue. I have only tried this branch https://github.com/malban/mapviz/tree/fix-multires-image-scale-kinetic, as I use ROS Kinetic.

Thank you for solving the issue.

PR #590, opened 2018-07-27
Right_Click_Services

alex94263:
We developed a new plugin, which is called right_click_services.
Upon right clicking into mapviz-plugin a service "available_commands" is called, which retreives commands/services available (passing the GPS coordinates as well).
By Choosing a command in the context-menu a user can choose one of the returned commands which then gets passed to the service "execute_commands".
We developed this plugin to tell our drone to move to a specific point, to specify a waypoint-route and let the drone fly along this route.
We feel like it is an improvement and therefore would like to see it in the main repository included.


evenator:
You appear to have committed a lot of extraneous files by accident.

pjreed (changes requested):
Thanks for the work; sorry it's taken me a long time to get to this, it's been a while since I've had the time to work on mapviz.  Other than the comments I've left on the diffs, it would also be good to have a description of this in the `README.md` and maybe a few examples of where it is useful.

alex94263:
@pjreed 
I tried to apply the changes requested, but I seem to be running  into problems with the Travis CI build failing.
I am not completely sure how to resolve these issues.

danthony06:
Closing this because of the age and the unresolved issues that are still on ROS 1.

PR #593, opened 2018-10-24
Create Coordinate Picker plugin

jatowler:
Listens for mouse clicks on the map, transforms them to a specified
frame, and prints a list of them in the plugin GUI. The most recent
coordinates are copied to the clipboard.


mapviz_plugins/src/coordinate_picker_plugin.cpp
```diff
@@ -0,0 +1,223 @@
+// *****************************************************************************
+//
+// Copyright (c) 2016, Southwest Research Institute® (SwRI®)
```
malban: might as well make the date 2018

agyoungs:
For consistency with the rest of mapviz, can the order of the latitude and longitude coordinates be swapped for the ws84 frame?  Currently the coordiate picker spits it out in (lon, lat) ordering.  It appears the rest of mapviz shows (lat, lon) ordering.

jatowler:
The picker uses (lon, lat) because it works like (x,y). But it’s easily changed if others agree.

pjreed:
If it displays (lat, lon) everywhere else, I'd probably rather use that, even though I always get confused when something stuffs WGS84 coordinates into a data structure with an "x" and a "y" and I have to remember that in that case it's (lon, lat).

agyoungs:
The only other thing that I noticed that might be nice is a clear button to clear the coordinates.  That's not a big deal to me though.

jatowler:
It turns out the order of display is defined by the order returned from `TransformManager`. I can add an exception for `wgs84` to reverse the order, but that seems fragile. Is that the right way to go about it?

PR #594, opened 2018-10-24
Create Coordinate Picker plugin

jatowler:
Listens for mouse clicks on the map, transforms them to a specified
frame, and prints a list of them in the plugin GUI. The most recent
coordinates are copied to the clipboard.

This is just the Indigo version of #593


pjreed:
No need to merge this in separately, it's in the new combined master branch.

PR #595, opened 2018-10-25
Get indigo version working on kinetic as well.

matt-attack:
Requires the new qt_metapackages.

### DIFF
--- a/mapviz/CMakeLists.txt
+++ b/mapviz/CMakeLists.txt
@@ -17,7 +17,36 @@ find_package(catkin REQUIRED COMPONENTS
 
 find_package(Boost REQUIRED system filesystem)
 find_package(OpenCV 2 REQUIRED)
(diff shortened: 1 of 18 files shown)


matt-attack:
You'll probably have to walk me through that

pjreed:
Ok, I think this is probably ready to go.

@malban , if you'd like to review this:

These changes should make a single branch of the mapviz repo that will build and use the right version of Qt on every supported ROS distro (indig/kinetic/lunar/melodic).  The current plan is:
1. Merge this into `indigo-devel`
2. Create a new branch off of `indigo-devel` named `master`
3. Merge `kinetic-devel` into `master` to make sure there are no changes in `kinetic-devel` we're missing; this will probably be messy and require some cleanup
4. Make `master` the default branch and deprecate both `indigo-devel` and `kinetic-devel`
5. Release new versions of mapviz for every ROS distro that use `master`
6. Never have to cherry-pick commits into separate devel branches ever again

PR #596, opened 2018-11-09
Lock canvas movement

austindodson:
Add variable for locking canvas on a mouseMove event.


mapviz_plugins/src/canvas_click_filter.cpp
```diff
@@ -36,7 +36,7 @@ namespace mapviz_plugins
 {
   CanvasClickFilter::CanvasClickFilter() :
     is_mouse_down_(false),
-    max_ms_(Q_INT64_C(500)),
+    //max_ms_(Q_INT64_C(500)),
```
pjreed: Why remove the default value for this variable?

pjreed:
I think this build failure was a result of changes I recently made to combine the Indigo and Kinetic branches.  Try merging the latest `master` into your branch and put it again and see if it works now.

PR #598, opened 2018-12-18
Add Measuring Plugin

matt-attack:
This adds the measuring plugin which lets the user measure distances between clicks in any frame.


matt-attack:
@jgassaway Would probably like to review this

PR #615, opened 2019-02-13
Set is_ready_ flag to true for non-Bing sources

### LINKED ISSUE #614: tile_map plugin doesn't work for anything *other* than Bing
A side effect of a recent fix to the `tile_map` plugin to improve Bing support accidentally broke all of the other map servers.

pjreed:
Fixes #614


KindDragon:
Please release new ASAP with these changes. Bug #614 create real problems for us

PR #617, opened 2019-02-15
[mapviz_plugins/attitude_indicator] Minor refactoring

KindDragon:
and redundant logging removed


pjreed:
Thanks, that looks good!

Issue #616, opened 2019-02-13
Add reset button to textured marker plugin


jbdaniel18:
Need to add reset button and clear marker function

Issue #611, opened 2019-02-05
Add alpha control to TexturedMarker plugin


matt-attack:
Sometimes it is hard to see a bunch of TexturedMarker layers at once. It would be nice to be able to specify an alpha multiplier for each layer at the plugin level.

This could be implemented by adding a number picker and multiplying the alpha value for each image here: https://github.com/swri-robotics/mapviz/blob/master/mapviz_plugins/src/textured_marker_plugin.cpp#L454

PR #625, opened 2019-03-14
Textured Marker Adjustments (#611, #616)

mattrich37:
Modified textured marker plugin for issue #611 by adding a slider bar for adjusting alpha of the selected textured marker set.

Fixed issue #616 by modifying ClearHistory() prototype and linking button to clear event.


mapviz_plugins/include/mapviz_plugins/textured_marker_plugin.h
```diff
@@ -94,12 +92,16 @@ namespace mapviz_plugins
...
   protected Q_SLOTS:
+    void SetAlphaLevel(int alpha);
     void SelectTopic();
     void TopicEdited();
+    void ClearHistory();
     void ProcessMarker(const marti_visualization_msgs::TexturedMarkerConstPtr marker);
     void ProcessMarkers(const marti_visualization_msgs::TexturedMarkerArrayConstPtr markers);
 
   private:
+    float alphaVal;
```
pjreed: Member variables should end with a `_`; e.g., `alphaVal_`

mapviz_plugins/src/textured_marker_plugin.cpp
```diff
@@ -71,8 +71,13 @@ namespace mapviz_plugins
     p3.setColor(QPalette::Text, Qt::red);
     ui_.status->setPalette(p3);
 
+    // Initialize the alpha value to default
+    alphaVal = 1.0f;
```
pjreed: If this is meant as a default value for this member variable, it'd be better for it to be part of the initializer list at the top of the constructor.

mapviz_plugins/src/textured_marker_plugin.cpp
```diff
@@ -95,9 +100,32 @@ namespace mapviz_plugins
 
   void TexturedMarkerPlugin::ClearHistory()
   {
+    ROS_INFO("Marker Clear all");
```
pjreed: This is just personal preference, but when I have a log statement that is only used to show every time a function is called, I prefer for it to print out the name of the function; e. g.,
```cpp
ROS_INFO("TexturedMarkerPlugin::ClearHistory()");
```
That helps to disambiguate it from any other logging statement that might be related to clearing markers, and if somebody else later sees this in the output and is trying to track down where it came from, it makes it more obvious.

mapviz_plugins/src/textured_marker_plugin.cpp
```diff
@@ -444,13 +472,16 @@ namespace mapviz_plugins
   {
     ros::Time now = ros::Time::now();
 
+    float _alphaVal = alphaVal; // Set all markers to same alpha value
```
pjreed: Why make a copy of this member variable here?

mattrich37: I wanted to copy the current value (at draw time) of the Alpha slider before iterating through the markers.  This should ensure that each marker in the list is set to the same alpha value for every draw step.

mapviz_plugins/src/textured_marker_plugin.cpp
mattrich37: I agree this should be a more specific message.  However, note that the text here is currently also displayed on the GUI under the 'Status' tagline.

pjreed: Are you sure?  I would only expect that if you're calling `PrintInfo`; `ROS_INFO` shouldn't change anything in the GUI.

mattrich37: Seems like it does not anymore.  Must have only happened in testing.

mapviz_plugins/src/textured_marker_plugin.cpp
pjreed: If the concern is that this value could be modified while `Draw` is running, as it is right now, I think that shouldn't be a concern; Qt slots are always called from Qt's main thread, and drawing is also always done on the main thread.  It looks like `alphaVal` is only being modified when the `SetAlphaLevel` slot is called, so it should never be modified at the same time that this drawing is happening.

Although there's also no harm in leaving this as-is in case it could be modified by another thread in the future.

Issue #610, opened 2019-02-05
Saving config doesnt add the extension


matt-attack:
When you save a new config in mapviz it does not add the mvc extension to the file name unless you manually type it in.

PR #626, opened 2019-03-14
Specify default configuration extension on save

mattrich37:
Fixes issue #610 by specifying default extension in QFileDialog for config saves


pjreed:
It seems like this PR also has several changes related to textured markers in it; did you mean for those to be included?

Issue #630, opened 2019-05-03
Fixed Frame selection field has focus set on start up


mattrich37:
When using the default mapviz launch configuration, the Fixed Frame selection field begins in focus with the cursor at the end of default frame.  Certain key press, such as a space or alphanumeric character, will cause the fixed frame to change.  This can be a nuisance when attempting to launch multiple files and the mapviz window intercepts a key press.

PR #631, opened 2019-05-03
Set main window as in focus on start #630

mattrich37:
Fixes issue #630 by setting the main window as the focused object.  This prevents other widgets from gaining focus by default, such as the Fixed Frame input field.


daniel-dsouza:
looks good to me. merging.

PR #635, opened 2019-06-14
Fixed marker plugin to use swri_transform_util

agyoungs:
This allows markers published in the wgs84 frame to work as expected.


matt-attack:
Great! @nick-alton this could fix all of our problem

PR #636, opened 2019-06-27
Add ability to show and hide markers by namespace

matt-attack:
image

### DIFF
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -99,6 +99,8 @@ namespace mapviz_plugins
     {
       initialized_ = false;
       markers_.clear();
+      marker_visible_.clear();
+      ui_.nsList->clear();
       has_message_ = false;
       PrintWarning("No messages received.");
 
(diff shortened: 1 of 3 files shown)


mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -405,6 +417,18 @@ namespace mapviz_plugins
 
   void MarkerPlugin::Draw(double x, double y, double scale)
   {
+    for (size_t i = 0; i < ui_.nsList->count(); i++)
```
pjreed: Any reason for this to not be `for (const auto& item : ui_.nsList)`?

pjreed:
Aside from my one comment, this looks good.  You can probably also remove the reference to `lunar` from the `.travis.yml` script.

mapviz_plugins/src/marker_plugin.cpp
pjreed: Nevermind that, I checked the class and discovered that it doesn't actually provide access to its container of items.  Lame.

PR #640, opened 2019-07-18
638 add text to measuring plugin

mdgrogan:
Add ability to display measurement tags alongside measurement lines on main map. Tags can be edited for color, font size, and transparency.

### DIFF
--- a/mapviz_plugins/include/mapviz_plugins/measuring_plugin.h
+++ b/mapviz_plugins/include/mapviz_plugins/measuring_plugin.h
@@ -55,6 +55,7 @@ namespace mapviz_plugins
       bool Initialize(QGLWidget* canvas);
       void Shutdown() { };
 
+      void Paint(QPainter* painter, double x, double y, double scale);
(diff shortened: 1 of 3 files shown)


mapviz_plugins/include/mapviz_plugins/measuring_plugin.h
```diff
@@ -31,6 +31,7 @@
 #define MAPVIZ_PLUGINS_MEASURING_PLUGIN_H_
 
 #include <mapviz/mapviz_plugin.h>
+#include <mapviz_plugins/point_drawing_plugin.h>
```
pjreed: Why is this being included here?

mdgrogan: It shouldn't be. Should I remove and re-commit?

pjreed: Yeah, please do.

PR #641, opened 2019-07-23
Add mapviz plug-in for PoseStamped messages.

malban:
Adds mapviz plug-in for PoseStamped messages.  This leverages the PointDrawingPlugin base class and is very similar to the other plugins like Odometry and GPS


pjreed:
Not sure why the Indigo build for this failed, but after hitting retry a few times, it worked.  If it keeps failing, I'll probably just remove Indigo since it's been EOL'ed anyway...

PR #643, opened 2019-07-23
Restore GL_UNPACK_ALIGNMENT to 4 to prevent corruption of Qt font rendering.

malban:
For some reason setting GL_UNPACK_ALIGNMENT to something other than 4 can effect Qt font rendering: https://forum.qt.io/topic/90149/qpainter-drawtext-display-corrupted-text-on-a-qopenglwidget

This PR restores GL_UNPACK_ALIGNMENT back to 4 in various places it is modified in the plug-ins after the relevant data has been unpacked.

Before: 
mapviz_20190723T204911_772000

After:
mapviz_20190723T204707_874595


pjreed:
Ooh, weird.  Good catch!

PR #646, opened 2019-09-12
Fix plug-in config panel to scroll per-pixel instead of per-item.

malban:
Fix plug-in config panel to scroll per-pixel instead of per-item, allowing configuration of plug-ins that have config settings that are too large to fit in the panel all at once.


malban:
Thanks!

pjreed:
No problem, this looks like a handy change.

PR #658, opened 2019-11-15
Add ability to use relative paths for images in the robot image plugin

matt-attack:
Add ability to use relative paths for images in the robot image plugin using the $(find x) syntax from launch files

### DIFF
--- a/mapviz_plugins/src/robot_image_plugin.cpp
+++ b/mapviz_plugins/src/robot_image_plugin.cpp
@@ -296,9 +296,11 @@ namespace mapviz_plugins
       }
 
       std::string real_filename;
-      if (filename_.find("$(find ") != -1)
+      size_t spos = filename_.find("$(find ");
+      bool has_close = spos != -1 ? filename_.find(')', spos) != -1: false;
+      if (spos != -1 && spos + 7 < filename_.size() && has_close)
       {
-        std::string package = filename_.substr(filename_.find("$(find ") + 7);
+        std::string package = filename_.substr(spos + 7);
         package = package.substr(0, package.find(")"));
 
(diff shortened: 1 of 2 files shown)


matt-attack:
@pjreed looks like the build timed out, causing it to fail

mapviz_plugins/src/robot_image_plugin.cpp
```diff
@@ -287,7 +295,21 @@ namespace mapviz_plugins
         texture_loaded_ = false;
       }
 
-      if (image_.load(filename_.c_str()))
+      std::string real_filename;
+      if (filename_.find("$(find ") != -1)
+      {
+        std::string package = filename_.substr(filename_.find("$(find ") + 7);
```
pjreed: What's the `+ 7` here for?  Is it safe to just got into an arbitrary index in the string like that?

matt-attack: It's safe but apparently it can throw an exception. I'll revise it to check the range.

matt-attack: Should be exception free now

danthony06: I think having at least a comment in there to explain where the +7 comes from would be helpful, or do a strlen() on that substring. Otherwise it's pretty hard to understand where that arbitrary number comes from.

pjreed: Also doesn't look like anything has changed, did you push?

matt-attack: I apparently pushed to the wrong origin

PR #663, opened 2019-12-03
Change for loop to while loop.

mdgrogan:
Change for loop to a while loop so that we don't skip items and overshoot the end of the container when erasing.

### DIFF
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -431,7 +431,8 @@ namespace mapviz_plugins
 
     ros::Time now = ros::Time::now();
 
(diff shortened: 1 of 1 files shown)


matt-attack:
I'm not sure that I see a difference in what is executed here. Does this fix a crash?

danthony06:
It looks like with the change it's possible to dereference an invalid
iterator.

mdgrogan:
Where does that occur?

danthony06:
Sorry about that, I misread the code on my phone and thought the continue
statements had been removed. I'm still not sure what problem this change is
addressing.

mdgrogan:
Ah, no worries. The erase function returns the next iterator and at the end of the for loop, that iterator will be incremented again. So it potentially skips items and increments the iterator past end.

PR #665, opened 2020-01-03
Add float plugin

### LINKED ISSUE #664: Add Speedometer
Would be nice to be able to visualize the vehicles speed in mapviz so one doesn't have to manually rostopic echo this. This could be grabbed from an odom message or a marti_sensor_msgs/Velocity.

matt-attack:
Somewhat resolves #664 

Adds a plugin that can render floats and marti_sensor_msgs/Velocity messages.


KindDragon:
Maybe you can also add a description and screenshot to README.md

matt-attack:
Good idea. Added a description. Didn't add a screenshot as there isn't much of a standard look to it, other than the config page.

PR #666, opened 2020-01-03
Make string plugin also support string stamped

matt-attack:


mapviz_plugins/src/string_plugin.cpp
```diff
@@ -387,9 +387,19 @@ namespace mapviz_plugins
     offset_y_ = offset;
   }
 
-  void StringPlugin::stringCallback(const std_msgs::StringConstPtr& str)
+#define IS_INSTANCE(msg, type) \
+  (msg->getDataType() == ros::message_traits::datatype<type>())
```
daniel-dsouza: You should probably check with @pjreed if he wants this as a macro

pjreed: While this isn't the end of the world since it's at least not in a header, but is there any reason for this to not be a function?

matt-attack: I guess not. It could be a templated function. My excuse is everyone else is doing exactly this.

matt-attack:
@pjreed I accidentally pushed this to origin. Is there a way to delete that branch?

matt-attack:
> apart from the macro, this looks good to m e.

@daniel-dsouza should be good now

pjreed:
> @pjreed I accidentally pushed this to origin. Is there a way to delete that branch?

Sure, done.

daniel-dsouza:
@pjreed looks good to merge

PR #667, opened 2020-01-11
Remove OpenGL warning

daniel-dsouza:
Uses GLVND libraries instead of legacy libraries

```
CMake Warning (dev) at /usr/share/cmake-3.15/Modules/FindOpenGL.cmake:275 (message):
  Policy CMP0072 is not set: FindOpenGL prefers GLVND by default when
  available.  Run "cmake --help-policy CMP0072" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  FindOpenGL found both a legacy GL library:

    OPENGL_gl_LIBRARY: /usr/lib/x86_64-linux-gnu/libGL.so

  and GLVND libraries for OpenGL and GLX:

    OPENGL_opengl_LIBRARY: /usr/lib/x86_64-linux-gnu/libOpenGL.so
    OPENGL_glx_LIBRARY: /usr/lib/x86_64-linux-gnu/libGLX.so

  OpenGL_GL_PREFERENCE has not been set to "GLVND" or "LEGACY", so for
  compatibility with CMake 3.10 and below the legacy GL library will be used.
Call Stack (most recent call first):
  /usr/share/cmake-3.15/Modules/FindGLU.cmake:12 (include)
  CMakeLists.txt:81 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This branch needs testing with more setups.


pjreed:
The build immediately fails:
```
Errors     << mapviz:cmake /root/catkin_ws/logs/mapviz/build.cmake.000.log

CMake Error at /root/catkin_ws/src/mapviz/mapviz/CMakeLists.txt:71 (cmake_policy):

  Policy "CMP0072" is not known to this version of CMake.
```

What's up with that?

PR #668, opened 2020-01-20
Remove extension test

matt-attack:
Not ready yet, just want to be able to run the CI.

Theoretically this fixes the ARM64 builds, or at least the OpenGL issue with them.

### DIFF
--- a/mapviz/src/map_canvas.cpp
+++ b/mapviz/src/map_canvas.cpp
@@ -33,6 +33,12 @@
 #include <cmath>
 #include <swri_math_util/constants.h>
 
+#ifndef __aarch64__
+#include <GL/glew.h>
+#include <GL/gl.h>
+#include <GL/glu.h>
+#endif
+
 namespace mapviz
 {
 
@@ -89,10 +95,12 @@ MapCanvas::MapCanvas(QWidget* parent) :
 
 MapCanvas::~MapCanvas()
 {
+#ifndef __aarch64__
   if(pixel_buffer_size_ != 0)
   {
-    glDeleteBuffers(2, pixel_buffer_ids_);
+    glDeleteBuffersARB(2, pixel_buffer_ids_);
   }
+#endif
 }
 
 void MapCanvas::InitializeTf(boost::shared_ptr<tf::TransformListener> tf)
@@ -100,12 +108,9 @@ void MapCanvas::InitializeTf(boost::shared_ptr<tf::TransformListener> tf)
   tf_ = tf;
 }
 
-void* (*glMapBuffer)(GLenum target, GLenum buffer);
-
-void (*glUnmapBuffer)(GLenum target);
-
 void MapCanvas::InitializePixelBuffers()
 {
+#ifndef __aarch64__
   if(has_pixel_buffers_)
   {
     int32_t buffer_size = width() * height() * 4;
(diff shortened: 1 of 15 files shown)


matt-attack:
@pjreed  how hard would it be to make the CI also build this in ARM64?

pjreed:
I've looked into it a little bit, and I don't think it'd be too hard to make the ROS Industrial scripts also run in an ARM64 image.  I'll experiment with it.

pjreed:
Ok, I think I've gotten the CI to build on ARM64 with both Kinetic and Melodic.  Try cherry-picking this into your branch: https://github.com/pjreed/mapviz/tree/add-arm64-ci

pjreed:
Drat, job timed out.  This should also be running in the PR tester on the ROS build farm, though, so I'm going to experiment and figure why that's not showing up here...

pjreed:
The mapviz config in the rosdistro repository was pointed at the `kinetic-devel` branch rather than `master`, which means that it's only testing PRs that target `kinetic-devel`.  I made a PR to fix that, but after it gets accepted, we'll have to wait overnight for it to reconfigure the Jenkins job on the build farm.

After that, the official ROS build farm will start building PRs... although it seems like it always builds them for amd64 rather than ARM, so that might not help us.

pjreed:
I added another commit to my branch that should work much better; it will make the builds run in native ARM64 build environments instead of qemu.  Give that a try.

matt-attack:
@danthony06  were you building on 18.04? Seems like the CI arm build fails in the exact same way on that but works on kinetic.

danthony06:
It was 18.04.

pjreed:
The ROS build farm also succeeds on 16.04 but fails on 18.04, so that seems right:
http://build.ros.org/job/Kbin_uxv8_uXv8__mapviz__ubuntu_xenial_arm64__binary/
http://build.ros.org/job/Mbin_ubv8_uBv8__mapviz__ubuntu_bionic_arm64__binary/

mapviz/src/map_canvas.cpp
```diff
@@ -28,9 +28,9 @@
 // *****************************************************************************
 
 
-#include <GL/glew.h>
-#include <GL/gl.h>
-#include <GL/glu.h>
+//#include <GL/glew.h>
```
pjreed: Probably  better to remove these (and other similar lines) instead of just commenting them out.

pjreed:
Nice, build looks good.

mapviz/src/map_canvas.cpp
matt-attack: Will do

matt-attack: Need to fix some other things that were commented out for testing as well

pjreed:
It looks like the build is now broken on AMD64 platforms: http://build.ros.org/job/Kpr__mapviz__ubuntu_xenial_amd64/9/console

matt-attack:
Should be fixed now

matt-attack:
Guess just needs to be tested on AMD64 to make sure I didnt break the screenshot acceleration

danthony06:
Looks like it's not working on a real display. When I try to add a point cloud or image, I get the following error on the command line, and mapviz appears to hang.

```bash
(mapviz:7969): GLib-GObject-WARNING **: 15:18:14.497: cannot register existing type 'GtkStyle'

(mapviz:7969): GLib-GObject-WARNING **: 15:18:14.497: cannot add private field to invalid (non-instantiatable) type '<invalid>'

(mapviz:7969): GLib-CRITICAL **: 15:18:14.497: g_once_init_leave: assertion 'result != 0' failed

(mapviz:7969): Gtk-CRITICAL **: 15:18:14.497: gtk_paint_vline: assertion 'GTK_IS_STYLE (style)' failed
```

danthony06:
Closing this in favor of: https://github.com/swri-robotics/mapviz/pull/871.

PR #675, opened 2020-04-04
Fix linking to OpenGL GLU library

lopsided98:
The `GLU_LIBRARY` variable was set by the deprecated `FindGLU.cmake` script. This package now uses `FindOpenGL.cmake`, which sets `OPENGL_glu_LIBRARY`.


pjreed:
Looks good, thanks!

PR #676, opened 2020-04-15
Increase odom queue size so messages dont drop

matt-attack:


pjreed:
Would it make sense to do this for the other point drawing plugins as well?  (GPS, Route, Tf, I forget if there are others)

matt-attack:
It probably would. I'll go ahead and do that

matt-attack:
I pushed that now. I went and changed all the single point types ones. I hesitate doing it for routes as those really shouldnt run into the same cases.

PR #679, opened 2020-04-27
Disable coordinate picker and measuring plugin when hidden (#678)

pjreed:
Cherry-picking this into the ROS2 branch.

### DIFF
--- a/mapviz_plugins/src/measuring_plugin.cpp
+++ b/mapviz_plugins/src/measuring_plugin.cpp
@@ -36,11 +36,7 @@
 #include <QTextStream>
 #include <QPainter>
 
-#if QT_VERSION >= 0x050000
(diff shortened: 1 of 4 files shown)


mapviz_plugins/src/measuring_plugin.cpp
```diff
@@ -36,11 +36,7 @@
 #include <QTextStream>
 #include <QPainter>
 
-#if QT_VERSION >= 0x050000
```
jgassaway, line 39: Are these unneeded in ros2?

pjreed: Yep, ROS2 (or more specifically, the distros ROS2 runs on) only supports Qt 5, so support for lower versions is just wasting space.

PR #680, opened 2020-05-08
Fix plan route visibility

matt-attack:
This disables the event filter and stops replanning when hidden. It was stealing clicks from other plugins and causing extra load when disabled.


pjreed:
Assuming this also exists in the `dashing-devel` branch, could you also cherry-pick this into it and open a PR for that?

matt-attack:
Sure. Man, there needs to be an easier way to do this.

pjreed:
Cool, thanks.

PR #682, opened 2020-05-10
Adding rotation to image_plugin

TobinHall:
Add ability to rotate images in steps of 90 degrees.

### DIFF
--- a/mapviz_plugins/src/image_plugin.cpp
+++ b/mapviz_plugins/src/image_plugin.cpp
@@ -222,19 +222,19 @@ namespace mapviz_plugins
   void ImagePlugin::SetRotation(QString rotation)
   {
 
-    if (rotation == "90°")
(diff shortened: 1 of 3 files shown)


danthony06 (approved):

mapviz_plugins/src/image_plugin.cpp
```diff
@@ -659,6 +703,20 @@ namespace mapviz_plugins
     return units_string;
   }
 
+  std::string ImagePlugin::RotationToString(int rotation)
+  {
+    std::string rotation_string = "0°";
```
matt-attack, line 708: I'd probably remove the degree characters since they are difficult to type manually.

danthony06: I made the degree symbol optional

PR #686, opened 2020-05-27
Add docs site

jatowler:
Creates a documentation site using GitHub Pages and the [Just the Docs](https://github.com/pmarsceill/just-the-docs) theme.

This PR removes the `doc` directory (moving the contents to `docs/assets/images` instead).

Note: after accepting, Line 4 of `_config.yml` will need to be updated to `https://swri-robotics.github.io` or none of the assets will load.

Also, GitHub Pages will need to be enabled using the `docs` directory.


docs/_config.yml
```diff
@@ -0,0 +1,50 @@
+title: "Mapviz"
+description: Modular ROS visualization tool for 2D data
+baseurl: "/mapviz" # the subpath of your site, e.g. /blog
+url: "https://jatowler.github.io" # the base hostname & protocol for your site, e.g. http://example.com
```
pjreed, line 4:
```suggestion
url: "https://swri-robotics.github.io" # the base hostname & protocol for your site, e.g. http://example.com
```
Might as well go ahead and make this change now; if you do more work on this, you'll want to create a new branch off of master, anyway.

PR #692, opened 2020-06-01
Use higher precision in the coordinate picker for wgs84

matt-attack:


mapviz_plugins/src/coordinate_picker_plugin.cpp
```diff
@@ -170,7 +170,14 @@ bool CoordinatePickerPlugin::handleMousePress(QMouseEvent* event)
   ROS_DEBUG("Transformed point in frame '%s': %f %f", frame.c_str(), point.x(), point.y());
   QString new_point;
   QTextStream stream(&new_point);
-  stream.setRealNumberPrecision(4);
+  if (frame == "/wgs84" || frame == "wgs84")
```
pjreed, line 173:
```suggestion
  if (swri_transform_util::FrameIdsEqual(frame, swri_transform_util::_wgs84_frame))
```
Just to avoid putting any hard-coded strings in here.

matt-attack:
Oh cool, was thinking there was a function for that but didn't know where it was.

PR #696, opened 2020-06-15
Support ROS Noetic

pjreed:
This makes tweaks necessary for mapviz to build and run on Ubuntu 20.04 /
ROS Noetic.  It also drops support for Ubuntu 14.04 / ROS Indigo.

Signed-off-by: P. J. Reed <preed@swri.org>


pjreed:
There's nothing complex here, this is just a good indicator of the kind of stuff that has to be changed when moving to Noetic.  We can drop support for Qt 4 / OpenCV 2 / libboost < 1.58 since those were for Ubuntu 14.04 / ROS Indigo, and they've been EOL'd for a while now.

CI build is currently failing in Noetic due to an upstream issue with swri_math_util;  I'm working on it, nothing here should have to change.

PR #699, opened 2020-06-16
Modern cmake updates

jhdcs:
PR for updating CMakeLists.txt files towards Modern CMake standards


mapviz/CMakeLists.txt
```diff
@@ -1,14 +1,25 @@
...
+  cmake_policy(SET CMP0072 NEW)
+endif()
+
 if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
-project(mapviz)
+project(mapviz 
+  VERSION 2.0.0
```
pjreed, line 20: I don't remember seeing a version set here in other ROS projects, is this a normal thing now?  I feel like this could be a little annoying if we have to also update this number every time we make a new release.

mapviz/CMakeLists.txt
```diff
@@ -122,6 +122,19 @@ add_library(rqt_${PROJECT_NAME}
...
+  PRIVATE
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_CURRENT_BINARY_DIR})
+target_include_directories(rqt_${PROJECT_NAME} SYSTEM
+  PUBLIC
+    ${Qt_INCLUDE_DIRS}
+    ${OpenGL_INCLUDE_DIR}
+    ${GLEW_INCLUDE_DIRS}
+    ${OpenCV_INCLUDE_DIRS}
+    ${YamlCpp_INCLUDE_DIR})
```
pjreed, line 137: We can probably clean up the include directories a little more; something to note is that if you use the modern syntax for specifying your `target_link_libraries` (as we do with `OpenGL::GL`, `Qt5::Concurrent`, etc. below), it will automatically add the header directories to the include path without needing to manually specify them.  So, we should at least be able to remove Qt and OpenGL from this list, and I believe there are also appropriate targets we can use for including OpenCV (https://stackoverflow.com/questions/55640570/using-opencvs-cmake-module-targets-instead-of-including-all-libraries-directly).  There might also be similar things we have use for GLEW or YamlCpp, I haven't looked them up.

mapviz/CMakeLists.txt
jhdcs: It's a recommended thing as far as the documentation for Modern CMake goes. However, if other ROS 2 projects don't include it, then we can leave it out. It's an optional parameter.

jhdcs: Removed the version numbers

mapviz/CMakeLists.txt
jhdcs: I have converted OpenGL_INCLUDE_DIR, GLEW_INCLUDE_DIRS, and OpenCV_INCLUDE_DIRS to their target equivalents. However, I'm not finding a target equivalent for YamlCpp. The generated targets cmake file appears to be in a CMake version that is too old for Modern CMake (<3.0)

jhdcs:
Re-ran Mapviz to make sure this is still running properly after all the CMake changes, and it looks like it's going fine. Can I get another review? See if there are some style changes you'd like to see?

pjreed (approved):

pjreed:
I was just looking over PRs and noticed this one had never been merged, although it looks fine to me.  Doing so now.

PR #700, opened 2020-06-18
Separate Save and Save As and open file dialogs in better locations.

matt-attack:


matt-attack:
@pjreed this looks good on my end. Should be good to merge if you get the chance to look it over.

mapviz/src/mapviz.ui
```diff
@@ -350,9 +351,9 @@
     <string>Open Config</string>
    </property>
   </action>
-  <action name="actionSave_config">
+  <action name="actionSaveAs_config">
    <property name="text">
-    <string>Save Config</string>
+    <string>Save Config As</string>
```
pjreed, line 356:
```suggestion
    <string>Save Config As...</string>
```

Common convention is to have menu options that open dialogs end with an ellipsis.

matt-attack: Somehow I was not aware of that. Interesting

pjreed:
Could you also cherry-pick this into `dashing-devel`?  Thanks!

PR #703, opened 2020-07-15
[tile_map] Output human-readable error message

kledom:
Rather than using the signal error(QNetworkReply::NetworkError), which does not expose the underlying QNetworkReply and only provides the QNetworkReply::NetworkError value, use the QNetworkReply::errorString() method to get a human-readable error message.

### DIFF
--- a/tile_map/src/image_cache.cpp
+++ b/tile_map/src/image_cache.cpp
@@ -202,8 +202,6 @@ namespace tile_map
         true);
 
     QNetworkReply *reply = network_manager_.get(request);
-    connect(reply, SIGNAL(error(QNetworkReply::NetworkError)),
-            this, SLOT(NetworkError(QNetworkReply::NetworkError)));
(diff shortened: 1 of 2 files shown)


kledom:
I'd also suggest to use ROS_ERROR_THROTTLE instead of ROS_ERROR, so that the logs are not flooded with error messages once the tile server becomes unavailable.

tile_map/src/image_cache.cpp
```diff
@@ -229,6 +227,7 @@ namespace tile_map
       }
       else
       {
+        ROS_ERROR("NETWORK ERROR: %s", reply->errorString().toStdString().c_str());
```
pjreed, line 230:
```suggestion
        ROS_ERROR_THROTTLE(1.0, "NETWORK ERROR: %s", reply->errorString().toStdString().c_str());
```

PR #704, opened 2020-09-10
Constrain the minimum line and point marker sizes to be 1 pixel wide.

malban:
Mapviz interprets the scale field of line and point markers as pixels instead of meters as specified in the Marker message.  There are some advantage to visualizing the line and point widths in mapviz in a zoom independent way.

However, this means that line and point markers don't display correctly in both mapviz and rviz (which does interpret the size as meters).

Since line and point markers with scale defined in meters for rviz usually have scales less than 1.0, they don't show up in mapviz very well.

This PR constrains the pixel size of line an point markers to be at least 1.0 in mapviz.

### DIFF
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -495,7 +495,7 @@ namespace mapviz_plugins
         glEnd();
       }
       else if (marker.display_type == visualization_msgs::Marker::LINE_STRIP) {
-        glLineWidth(marker.scale_x);
+        glLineWidth(std::max(1.0f, marker.scale_x));
         glBegin(GL_LINE_STRIP);
 
         for (const auto &point : marker.points) {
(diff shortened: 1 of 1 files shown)


mapviz_plugins/src/marker_plugin.cpp
```diff
@@ -523,7 +523,7 @@ namespace mapviz_plugins
         glEnd();
       }
       else if (marker.display_type == visualization_msgs::Marker::POINTS) {
-        glPointSize(marker.scale_x);
+        glLineWidth(std::max(1.0f, marker.scale_x));
```
matt-attack, line 526: Should this actually be point size?

malban: yep ...

pjreed:
Looks good to me, although it looks like the Noetic build has some warnings.  I didn't see those in the build on the master branch, are those new or is the build system just being flaky?

malban:
> Looks good to me, although it looks like the Noetic build has some warnings. I didn't see those in the build on the master branch, are those new or is the build system just being flaky?

I don't think anything in the changes should cause additional warnings.  Maybe the noetic build environment was updated causing it to have more verbose warnings>

pjreed:
Yeah, I don't think those cmake warnings are our fault, so I'm going to just ignore them.

PR #705, opened 2020-09-11
Restore and refactor use_latest_transforms option for position history plugins.

malban:
The `use_latest_transforms`  checkbox doesn't appear to work as intended since https://github.com/swri-robotics/mapviz/commit/bc5f5c5a9060357259528f22e71ec711df4d95e7.

It was originally intended to allow drawing the position history of a bot in the frame of the position topic, but rigidly transformed using the latest transform between the position topic frame and the fixed from of mapviz.   This makes it somewhat intuitive to see the divergence of different localization sources such as a relative state estimate and an absolute state estimate.

As it works now, like in rviz, the positions are converted to the fixed frame at the time they are received using the stamp of the position.  This results in eliminating the divergence of the position histories from the rendering, which is not very helpful.

This MR refactors the `use_latest_transforms` parameter to be per-plugin instead of a global setting.  The global setting was problematic because it only applied to position history plugins like Odometry, Tf, Pose, etc.   It also restores honoring that option to the position history plugins.

### DIFF
--- a/mapviz/include/mapviz/map_canvas.h
+++ b/mapviz/include/mapviz/map_canvas.h
@@ -71,7 +71,6 @@ namespace mapviz
     void ToggleFixOrientation(bool on);
     void ToggleRotate90(bool on);
     void ToggleEnableAntialiasing(bool on);
-    void ToggleUseLatestTransforms(bool on);
     void UpdateView();
     void ReorderDisplays();
     void ResetLocation();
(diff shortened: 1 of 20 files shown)


pjreed:
This looks good to me, could you check and see how much work it'd be to cherry-pick this into the ROS2 branch (`dashing-devel`)?

malban:
> This looks good to me, could you check and see how much work it'd be to cherry-pick this into the ROS2 branch (`dashing-devel`)?

Probably not very much.  I'll take a whack at it later in the week if I don't forget.  Is there a ROS2 docker image or something?

pjreed:
Yep, `ros:foxy` is the latest.  I haven't actually released mapviz into foxy yet, but all of its dependencies are there.

matt-attack (approved):

matt-attack:
If I don't forget, should be able to test and merge this sometime this week.

danthony06 (approved):

PR #706, opened 2020-09-11
Fix compile warning in ROS Noetic

pjreed:
A log statement was using the wrong format specifier for some size_t variables.

Signed-off-by: P. J. Reed <preed@swri.org>


pjreed:
The cmake warnings for the Npr builds are weird and I don't think they're our fault, so I'm going to merge this anyway.

PR #715, opened 2021-01-07
Add visualizer for PoseArray

knickels:
This PR adds a PoseArray visualizer plugin, modeled after the Pose visualizer.


mapviz_plugins/include/mapviz_plugins/pose_array_plugin.h
```diff
@@ -0,0 +1,104 @@
...
+// Include mapviz_plugin.h first to ensure GL deps are included in the right order
+#include <mapviz/mapviz_plugin.h>
+
+#include <mapviz/map_canvas.h>
+#include <mapviz_plugins/point_drawing_plugin.h>
+
+// C++ standard libraries
+#include <list>
+#include <string>
+#include <vector>
```
danthony06, line 43: Small nitpick, but are these includes actually needed?

mapviz_plugins/src/pose_array_plugin.cpp
```diff
@@ -0,0 +1,299 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) 
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE 
+ * POSSIBILITY OF SUCH DAMAGE.
+ **/ 
+
+#include <mapviz_plugins/pose_array_plugin.h>
+
+// C++ standard libraries
+#include <cstdio>
+#include <vector>
```
danthony06, line 35: Are these includes actually needed? The ```cstdio``` seems a little suspicious.

mapviz_plugins/src/pose_array_plugin.cpp
```diff
@@ -0,0 +1,299 @@
...
+      TopicEdited();
+    }
+  }
+
+  void PoseArrayPlugin::TopicEdited()
+  {
+    std::string topic = ui_.topic->text().trimmed().toStdString();
+    if (topic != topic_)
+    {
+      initialized_ = false;
```
danthony06, line 116: What is initialized_ used for? I see it assigned to, but we don't seem to check it anywhere.

mapviz_plugins/src/pose_array_plugin.cpp
knickels: Nope, holdovers from pose_plugin.cpp.  I'll remove.

mapviz_plugins/src/pose_array_plugin.cpp
knickels: I don't see initialized_ or has_message_ being used for anything, but I'm suspicious since they are in pretty much all plugins?  Are they vestigal?

pjreed: `initialized_` is used in the base MapvizPlugin class; Mapviz will not draw a plugin until that is true, so plugins can use that to disable rendering until they're ready.

`has_message_` isn't part of any base class, but many plugins use something like that to track whether they've received any messages or not, so it may just be a copy/paste issue here.

knickels:
Hmm, @pjreed does mapviz cache it's list of plugins somewhere?  I flipped back to the pose_array branch to test this, but now mapviz claims it can't see pose_array plugin.

mapviz_plugins/src/pose_array_plugin.cpp
```diff
@@ -0,0 +1,299 @@
...
+                                        pose.position.y,
+                                        pose.position.z);
+
+        stamped_point.orientation = tf::Quaternion(
+                pose.orientation.x,
+                pose.orientation.y,
+                pose.orientation.z,
+                pose.orientation.w);
+
+        pushPoint( std::move( stamped_point) );
```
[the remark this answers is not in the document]
matt-attack: It does seem that this could make it tricky to use as you would have to know how many poses are in each message if you planned on only showing one.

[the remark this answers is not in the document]
knickels: Yes, I noticed this too - currently removing the buffer option, as I don't think it makes too much sense for an array.  I don't see a great value in displaying the last N pose arrays :)

mapviz_plugins/src/pose_array_plugin.cpp
knickels: Turns out that this one uses it just like that (that part was just copied from pose_plugin, but not by error.)

knickels:
OK, @matt-attack @danthony06 @pjreed I think I've addressed everything.  Thanks for the feedback!

danthony06 (approved):

PR #716, opened 2021-01-15
Pose publisher

knickels:
This PR adds a plugin to publish a selected pose, similar to move_base but with a configurable pose and frame_id.


shrijitsingh99:
Any plans for this to get merged in or are there some blockers?

PR #718, opened 2021-01-23
Allow text markers to be enabled/disabled via their namespace using the marker plugin checkboxes

agyoungs:
Currently, the marker plugin only skips drawing markers if they're disabled via their namespace using the Draw function. Text markers are handled within the Paint function so this PR handles that using the same functionality found in the Draw function.


matt-attack (approved):
Looks good!

PR #721, opened 2021-02-21
720 allow building with c++14 for version greater than kinetic

### LINKED ISSUE #720: Allow building with C++14 for versions > ROS Kinetic
ROS Melodic and greater default builds to target C++14. I think all mapviz packages should default to those settings unless building for a version of ROS <= Kinetic, in which case builds should target C++11.

rjb0026:
Fixes #720

### DIFF
--- a/mapviz/CMakeLists.txt
+++ b/mapviz/CMakeLists.txt
@@ -95,12 +95,6 @@ catkin_package(
 # Fix conflict between Boost signals used by tf and QT signals used by mapviz
 add_definitions(-DQT_NO_KEYWORDS)
 
-# If using <= kinetic, require C++11, otherwise default to C++14
-if ($ENV{ROS_DISTRO} STREQUAL "kinetic" OR $ENV{ROS_DISTRO} STRLESS "kinetic")
-  set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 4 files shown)


mapviz/CMakeLists.txt
```diff
@@ -95,6 +95,12 @@ catkin_package(
 # Fix conflict between Boost signals used by tf and QT signals used by mapviz
 add_definitions(-DQT_NO_KEYWORDS)
 
+# If using <= kinetic, require C++11, otherwise default to C++14
+if ($ENV{ROS_DISTRO} STREQUAL "kinetic" OR $ENV{ROS_DISTRO} STRLESS "kinetic")
+  set(CMAKE_CXX_STANDARD 11)
```
matt-attack, line 100: Overall this looks okay. However, setting the C++ standard globally like this isn't the best. I would make sure its set per target like before instead.

rjb0026: Done. Not sure why those ubuntu focal and debian buster jobs are failing... doesn't look related to the build.

matt-attack (approved):
Looks good! I'll test this on my kinetic system to be sure, but should be good to go afterwards. Remind me if I dont get to it tomorrow.

PR #737, opened 2021-06-06
736 add a duplicate feature for plugins

### LINKED ISSUE #736: Add a "duplicate" feature for plugins
I've found that I will often need many of the same plugin that are all configured the same with the exception of one or two properties. It would be nice to have the ability to set up one plugin as "template" for the common properties, duplicate it a bunch of times, and then fix up the non common properties. Something like the "duplicate" feature like [the one added in RVIZ](https://github.com/ros-visualization/rviz/pull/977).

Right now I'm thinking about adding "duplicate" as an option in the drop down context menu that appears when you right click a plugin and adding the 'Ctrl+d' hotkey. I'm not opposed to adding a "duplicate" button next to the "add" and "remove" buttons either if others think that is useful.

rjb0026:
Closes #736. Added a "duplicate" feature to quickly create a deep copy of a plugin including all its configurations. Current methods for creating a duplicate:

- Right click on the plugin to duplicate and select the "duplicate" option in the drop down
- Select the plugin to duplicate and press the "Ctrl+d" hotkey

https://user-images.githubusercontent.com/33141599/120910447-b9126b80-c644-11eb-82f6-048480bb7eb5.mp4

### DIFF
--- a/mapviz/include/mapviz/mapviz.h
+++ b/mapviz/include/mapviz/mapviz.h
@@ -51,7 +51,6 @@
 #include <QWidget>
 #include <QStringList>
 #include <QMainWindow>
-#include <QShortcut>
 
 // ROS libraries
 #include <ros/ros.h>
--- a/mapviz/include/mapviz/mapviz.h
+++ b/mapviz/include/mapviz/mapviz.h
@@ -51,6 +51,7 @@
 #include <QWidget>
 #include <QStringList>
 #include <QMainWindow>
+#include <QShortcut>
 
 // ROS libraries
 #include <ros/ros.h>
@@ -92,6 +93,8 @@ namespace mapviz
     void SelectNewDisplay();
     void RemoveDisplay();
     void RemoveDisplay(QListWidgetItem* item);
+    void DuplicateDisplay();
+    void DuplicateDisplay(QListWidgetItem *item);
     void ReorderDisplays();
     void FixedFrameSelected(const QString& text);
     void TargetFrameSelected(const QString& text);
--- a/mapviz/src/config_item.cpp
+++ b/mapviz/src/config_item.cpp
@@ -43,7 +43,6 @@ namespace mapviz
 
     edit_name_action_   = new  QAction("Edit Name", this);
     duplicate_item_action_ = new QAction("Duplicate", this);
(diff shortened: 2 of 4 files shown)


danthony06:
Looks pretty useful at first glance. I need to look at the temp file code, that seems a little odd.

rjb0026:
There might be a more efficient way, but I was trying to use the existing plugin interface. I didn't initially see a super easy way to go from `YAML::Emitter` to `YAML::Node` directly. I did however see a way to go from `YAML::Node` to `YAML::Emitter` easily. It would be nice if the `MapvizPlugin::SaveConfig` interface took in a `YAML::Node` directly and then let the user decide how to save that since it's not actually saving to a file in any of the implementations I've looked at.

Changing the existing `MapvizPlugin` interface or adding another pure virtual method to the `MapvizPlugin` class would break all existing plugins. Avoiding that was my primary objective for doing something kinda quick and dirty. I can look around in yaml-cpp to see if there is a more efficient way to do that conversion in case I'm just missing something.

rjb0026:
@danthony06 I found a better way to save/load the config for the plugin to be duplicated that does not require using a temp file. I've tested locally on melodic. The CI seems to be failing because it can't find catkin tools, so It didn't even make it to the build.

mapviz/include/mapviz/mapviz.h
```diff
@@ -51,6 +51,7 @@
 #include <QWidget>
 #include <QStringList>
 #include <QMainWindow>
+#include <QShortcut>
```
matt-attack, line 54: Any particular reason this was included?

mapviz/src/config_item.cpp
```diff
@@ -42,10 +42,13 @@ namespace mapviz
     ui_.setupUi(this);
 
     edit_name_action_   = new  QAction("Edit Name", this);
+    duplicate_item_action_ = new QAction("Duplicate", this);
+    duplicate_item_action_->setShortcut(QKeySequence(Qt::CTRL | Qt::Key_D));
```
matt-attack, line 46: Adding a shortcut just for this single thing might be a bit confusing.

danthony06:
@rjb0026 we never got this merged in. If you want this in ROS 2, can you open a new PR for it?

rjb0026:
@danthony06 Yeah I'll try to find some time to port and test this.

PR #738, opened 2021-06-15
Update to ROS2 galactic

AdamGoertz:
Fix warnings and deprecations for ROS galactic.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,8 @@ jobs:
     strategy:
       matrix:
         env:
-          - {ROS_DISTRO: dashing,  ROS_REPO: testing}
-          - {ROS_DISTRO: dashing,  ROS_REPO: main}
-          - {ROS_DISTRO: eloquent, ROS_REPO: testing}
-          - {ROS_DISTRO: eloquent, ROS_REPO: main}
-          - {ROS_DISTRO: foxy,     ROS_REPO: testing}
-          - {ROS_DISTRO: foxy,     ROS_REPO: main}
+          - {ROS_DISTRO: galactic,  ROS_REPO: testing}
+          - {ROS_DISTRO: galactic,  ROS_REPO: main}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -7,12 +7,8 @@ jobs:
     strategy:
       matrix:
         env:
-          - {ROS_DISTRO: dashing,  ROS_REPO: testing}
-          - {ROS_DISTRO: dashing,  ROS_REPO: main}
-          - {ROS_DISTRO: eloquent, ROS_REPO: testing}
-          - {ROS_DISTRO: eloquent, ROS_REPO: main}
-          - {ROS_DISTRO: foxy,     ROS_REPO: testing}
-          - {ROS_DISTRO: foxy,     ROS_REPO: main}
+          - {ROS_DISTRO: galactic,  ROS_REPO: testing}
```
danthony06, line 10: I'm not sure we want to remove foxy yet, because that's still a supported ROS distro.

mapviz_plugins/src/textured_marker_plugin.cpp
```diff
@@ -197,7 +200,7 @@ void TexturedMarkerPlugin::ProcessMarker(const marti_visualization_msgs::msg::Te
       markerData.expire_time = rclcpp::Time::max();
     } else {
       // Temporarily add 5 seconds to fix some existing markers.
-      markerData.expire_time = rclcpp::Time() + lifetime + rclcpp::Duration(5);
+      markerData.expire_time = rclcpp::Time() + lifetime + rclcpp::Duration(5ns);
```
danthony06, line 203: The original functionality was 5 seconds, not 5 nanoseconds.

danthony06 (changes requested):

.github/workflows/main.yml
AdamGoertz: I'd agree, but some of the changes break compatibility with foxy. Specifically, rclcpp::Duration::from_nanoseconds() doesn't exist in foxy (according to the api documentation at least).

mapviz_plugins/src/textured_marker_plugin.cpp
AdamGoertz: And it was working as intended? Interesting, since that constructor takes a number of nanoseconds. Either way, I can fix that.

danthony06 (approved):

PR #742, opened 2021-08-31
731 Fix exported dependencies in multires_image (Foxy / dashing-devel)

### LINKED ISSUE #731: [Foxy] multires_image doesn't export dependencies
Looking at the CMakeLists.txt for multires_image a lot of packages are pulled in but the dependencies are exported are in a non initialized variable "RUNTIME_DEPS", and then Qt.

Is that variable filled elsewhere or we should update the RUNTIME_DEPS to include the packages that it finds? Also, It seems exporting "Qt" is not correct, should it at last be "Qt5" or the specific Qt5 packages that are there?

On a separate note, if it is still a problem in "swri_transform_util" with OpenCV where if you use components you should declare them in a CONFIG_EXTRAS: https://answers.ros.org/question/331089/ament_export_dependenciesboost-not-working/?answer=332460#post-id-332460. That can be a ticket created in the right repo though, just saw the comment when pulling in OpenCV here, but it might apply here since it looks like we are depending on components of Boost.

vknisley-swri:
I set the `RUNTIME_DEPS` variable and included the different, specific Qt5 packages within it. I also added a `CONFIG_EXTRAS` file for the Boost components. Please let me know if I need to make any further changes.

Closes #731


multires_image/CMakeLists.txt
```diff
@@ -79,9 +79,9 @@ ament_target_dependencies(${PROJECT_NAME}
   gps_msgs
   mapviz
   pluginlib
-  rclcpp 
+  rclcpp
```
danthony06, line 82: Small change, but as along as we're dropping the extraneous line endings, could we get these dependencies in the right alphabetical order?

multires_image/cmake/export_components.cmake
```diff
@@ -0,0 +1 @@
+find_package(Boost REQUIRED COMPONENTS filesystem thread)
```
danthony06, line 1: Can you add a line ending after this to make the linter happy?

vknisley-swri: I added a new line at the end of the file.

multires_image/CMakeLists.txt
vknisley-swri: The dependencies should now be listed in the correct order.

danthony06 (approved):

multires_image/CMakeLists.txt
```diff
@@ -163,7 +163,26 @@ install(DIRECTORY launch
...
+  swri_math_util
+  swri_transform_util
+  tf2
+)
+
+ament_export_dependencies(${RUNTIME_DEPS})
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME})
-ament_package()
+ament_package(CONFIG_EXTRAS cmake/export_components.cmake)
```
matt-attack, line 188: This seems like a really weird way to export this. Is this actually the recommended approach?

matt-attack: After research it is, but this is very silly.

PR #750, opened 2022-01-07
Update GPS subscrption to use sensor qos

shrijitsingh99:
All sensors are generally expected to use the sensor QoS. There is even a REP proposal to standardise it https://github.com/ros-infrastructure/rep/pull/212. Besides that one of the most commonly use packages for localization i.e. robot_localization also uses that [QoS](https://github.com/cra-ros-pkg/robot_localization/blob/5c9d411e3823859f240b7594de41cfe896327914/src/navsat_transform.cpp#L171), so it would make sensr for Mapviz to also support it.

### DIFF
--- a/mapviz_plugins/src/navsat_plugin.cpp
+++ b/mapviz_plugins/src/navsat_plugin.cpp
@@ -108,7 +108,7 @@ namespace mapviz_plugins
       {
         navsat_sub_ = node_->create_subscription<sensor_msgs::msg::NavSatFix>(
             topic_,
-            rclcpp::QoS(1),
+            rclcpp::SensorDataQoS(rclcpp::KeepLast(1)),
             std::bind(&NavSatPlugin::NavSatFixCallback, this, std::placeholders::_1));
 
         RCLCPP_INFO(node_->get_logger(), "Subscribing to %s", topic_.c_str());


danthony06:
@daniel-dsouza I think we need a pretty good review of this. Are you interested in taking a look at it?

shrijitsingh99:
Sorry, about that. I pushed the wrong commit before. This PR is just a minor change. Though it will require changes at multiple places since many of the plugins subscribe to sensor data.

shrijitsingh99:
Friendly ping
Just adding more information sensor data qos just makes a smaller queue size and set reliability to best effort. Having best efforts reliability give maximum compatibility since then it works with both reliable and best effort publishers.
https://docs.ros.org/en/rolling/Concepts/About-Quality-of-Service-Settings.html

matt-attack:
Since this is for visualization, there is actually value in having reliable transport unless the network doesn't permit it. Generally I would like to see **all** the messages and not just whatever happens to make it through for debugging purposes. Perhaps adding an option to mapviz to let the user specify what reliability they would like on a global basis would be a good idea.

danthony06:
We also should merge this into the ros2-devel branch.

PR #759, opened 2022-05-25
Check topic for type to determine which subscription callback to trigger

### LINKED ISSUE #756: unable to subscribe to MarkerArray under ROS 2 Galactic
I'm trying to use the Marker plugin with a Mapviz install under ROS 2 Galactic to display a MarkerArray, but am getting the following error:
```
[mapviz-1] [ERROR] [1648669993.058815599] [mapviz]: Unhandled RCLError in Qt event loop: could not create subscription: failed to create topic, at /tmp/binarydeb/ros-galactic-rmw-cyclonedds-cpp-0.22.4/src/rmw_node.cpp:2552, at /tmp/binarydeb/ros-galactic-rcl-3.1.2/src/rcl/subscription.c:105
```
I had an earlier installation of Mapviz under ROS 2 Eloquent that was able to subscribe to the same topic and messages I'm attempting to visualize under Galactic. I reran `colcon build` on the publisher package under Galactic and confirmed that the messages were publishing with the command `ros2 topic echo /poi_marker_array visualization_msgs/msg/MarkerArray`. Any thoughts on whether this is a problem with Mapviz or with my publisher, or additional troubleshooting tips?

agyoungs:
Closes #756

### DIFF
--- a/mapviz_plugins/src/marker_plugin.cpp
+++ b/mapviz_plugins/src/marker_plugin.cpp
@@ -123,14 +123,42 @@ namespace mapviz_plugins
     marker_array_sub_.reset();
     if (!topic_.empty())
     {
-      marker_sub_ = node_->create_subscription<visualization_msgs::msg::Marker>(
-          topic_,
-          rclcpp::QoS(100),
-          std::bind(&MarkerPlugin::handleMarker, this, std::placeholders::_1));
-      marker_array_sub_ = node_->create_subscription<visualization_msgs::msg::MarkerArray>(
-          topic_,
-          rclcpp::QoS(100),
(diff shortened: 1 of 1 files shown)


agyoungs:
@danthony06 Can you review this? The disadvantage to this approach is that if a topic doesn't exist when the plugin is initialized, the subscription will not be made. This could be addressed with a function override for plugins that gets triggered when new topics are identified.

The alternative approach is to use a generic subscription. However this will require de-serialization within the plugin.

danthony06 (approved):

danthony06:
Seems reasonable to me.

PR #765, opened 2022-10-10
Fix Plan Route plugin in ROS2 Humble

pjreed:
The Plan Route plugin was previously creating a publisher for the swri_route_util::Route class, which is not an actual ROS message but a wrapper around the marti_nav_msgs::Route message.  ROS2 Humble appears to have a compile-time check that prevents creating publishers for classes that are not ROS2 message definitions, so that will fail.

This changes the type of the publisher to marti_nav_msgs::Route and manually converts the message to that type before publishing it.

Signed-off-by: P. J. Reed <phillipreed@hatchbed.com>


pjreed:
After merging this and https://github.com/swri-robotics/marti_common/pull/693, mapviz should work on ROS2 Humble now.

danthony06 (approved):

PR #798, opened 2023-11-10
added action panel in addition to config panel

AnthonyZJiang:
## Overview
Added an action panel which allow plugins to be displayed in an additional panel.

## What I did not like before
The default UI hosts only one plugin panel. It can easily get very cluttered when a few plugins are added. It is difficult to navigate the list and locate the correct plugin to play with, especially on a mobile laptop with a small screen and a track pad. Frequently, misclicking could mess up the plugin list.

Screenshot from 2023-11-10 17-51-15

## Solution
My solution is to add a second panel called Action panel. In this panel, I can add the plugins such as move_base, draw_polygon etc. which I frequently use during field work. Then I will hide the Config panel, so the the fancy display components I configured earlier stay intact.

Screenshot from 2023-11-10 17-54-31

#### Visibility
Both panels can be toggled on and off through menu separately.
#### Config saving and loading
I have made sure save and open config functions behave correctly for both panels.

An additional property, `listName`, is added to each item under the display category to indicate their designated panel:

```
displays:
  - listName: config
    type: mapviz_plugins/laserscan
    name: new display
    config: ....
  - listName: action
    type: mapviz_plugins/attitude_indicator
    name: new display
    config: ....
```
Backward compatibility for old config files [@line 720](https://github.com/ICE9-Robotics/mapviz/blob/d3f23b07b0c48399fe12795b0214bbc4e09cffde/mapviz/src/mapviz.cpp#L720):
```
        std::string listName, type, name;
        if (displays[i]["listName"]) {
          displays[i]["listName"] >> listName;
        } else {
          listName = "config";
        }
        displays[i]["type"] >> type;
        displays[i]["name"] >> name;
```

#### Default panel location
The Config panel defaults to the left docking area, so the Action panel defaults to the right docking area.

### DIFF
--- a/mapviz/include/mapviz/mapviz.h
+++ b/mapviz/include/mapviz/mapviz.h
@@ -92,13 +92,14 @@ namespace mapviz
     void SelectNewDisplay();
     void RemoveDisplay();
     void RemoveDisplay(QListWidgetItem* item);
-    void ReorderDisplays();
+    void ReorderDisplays(std::string listName);
     void FixedFrameSelected(const QString& text);
     void TargetFrameSelected(const QString& text);
     void UpdateFrames();
     void SpinOnce();
     void UpdateSizeHints();
     void ToggleConfigPanel(bool on);
+    void ToggleActionPanel(bool on);
     void ToggleStatusBar(bool on);
     void ToggleCaptureTools(bool on);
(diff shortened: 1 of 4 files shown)


danthony06:
Looks interesting, let me try it out

PR #807, opened 2023-12-05
fixed button connect bug in move_base_plugin

tmxkn1:
Callback functions for `pushButtonInitialPose`, `pushButtonGoalPose` and `pushButtonAbort` are not linked correctly, causing the abort function not working and the intended cursor changes not firing.

Tested with Ubuntu 18.04, Melodic, master branch.


danthony06 (approved):
Looks good, thanks for the fix.

PR #818, opened 2024-04-11
WIP: Configurable qos

danthony06:
Updated the topic select UI to include QoS settings


danthony06:
@rjb0026 I don't have the ability to save the QoS parameters yet, and I haven't really tested this beyond getting it to compile.

mapviz/include/mapviz/mapviz_plugin.h
```diff
@@ -318,7 +318,38 @@ public Q_SLOTS:
...
+
+    if (node[prefix + "qos_durability"])
+    {
+      qos.durability = static_cast<rmw_qos_durability_policy_e>(node[prefix + "qos_durability"].as<int>());
+    }
+  }
+
+  void SaveQosConfig(YAML::Emitter& emitter, const rmw_qos_profile_t& qos, const std::string prefix = "") const
+  {
+    emitter << YAML::Key << prefix << "qos_depth" << YAML::Value << qos.depth;
```
rjb0026, line 346:
```suggestion
    emitter << YAML::Key << prefix + "qos_depth" << YAML::Value << qos.depth;
```
The `<<` operator was doing funky things here when I tried this out and concatenating the strings with just a `+` fixed it for me in local testing.

rjb0026 (changes requested):
Btw, I'm getting compiler warnings from plugins that include this file and others that have `Q_OBJECT;` after the `class ...` declaration because it is saying the ";" character is redundant. Can we go through and remove the ";" characters from the `Q_OBJECT;` lines?

PR #833, opened 2025-01-02
Fix topic lists

nick-alton:
Not 100% sure why, but I was testing that last change I made and noticed sometimes the displayed topic list and the actual topics lost sync. So you it would display names that when clicked would result in different topics being selected. The names were also incorrect for the topics. This change just allows the update logic to check the actual displayed names and use those for the basis of the update instead of the purely internally tracked logic it was using previously.


danthony06 (approved):

danthony06:
Sorry it took me so long to merge. I'd like to do it a different way, but that would probably trigger a much larger change.

PR #844, opened 2025-05-30
Add missing deps, remove ROS1-exclusive one, and fix include install location

vknisley-swri:
After making some of these changes, I was able to build my dependent package. I also added some other seemingly missing dependencies to the export list.

I also have my IDE set up to automatically remove trailing spaces, which led to some of the changes.

### DIFF
--- a/mapviz_plugins/CMakeLists.txt
+++ b/mapviz_plugins/CMakeLists.txt
@@ -103,10 +103,10 @@ set(SRC_FILES
   src/image_plugin.cpp
   src/laserscan_plugin.cpp
   src/marker_plugin.cpp
-  src/measuring_plugin.cpp 
+  src/measuring_plugin.cpp
   src/navsat_plugin.cpp
   src/occupancy_grid_plugin.cpp
-  src/odometry_plugin.cpp 
+  src/odometry_plugin.cpp
   src/path_plugin.cpp
   src/placeable_window_proxy.cpp
   src/plan_route_plugin.cpp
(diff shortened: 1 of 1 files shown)


mapviz_plugins/CMakeLists.txt
```diff
@@ -209,13 +209,13 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
...
     ${visualization_msgs_TARGETS})
-target_include_directories(${PROJECT_NAME} INTERFACE 
+target_include_directories(${PROJECT_NAME} INTERFACE
   ${swri_roscpp_TARGETS}
 )
 
 ### Install the plugins ###
 install(DIRECTORY include/
-  DESTINATION include/${PROJECT_NAME}
+  DESTINATION include/
```
danthony06: I really don't think we should change this. @DangitBen what are your thoughts here?

danthony06 (changes requested):
I think we need to keep the destination as include/${PROJECT_NAME}

mapviz_plugins/CMakeLists.txt
DangitBen: According to the `ament_cmake` [documentation](https://docs.ros.org/en/humble/How-To-Guides/Ament-CMake-Documentation.html#installing), installing to `include/${PROJECT_NAME}` is the correct location since humble. I think the reason is consistency between merged and isolated installs.
So in an isolated install the path might look like:
`workspaceDir/install/mapviz_plugins/include/mapviz_plugins/mapviz_plugins/some_header.h`
..............................................................................................^ paths in source are relative to this dir

The reason the project name appears twice is to keep things consistent when doing a merged-install. Which might generate a path like:
`workspaceDir/install/include/mapviz_plugins/include/mapviz_plugins/some_header.h`
......................................................................................^ paths in source are relative to this dir

You can see that if you don't include the last project name then the path in the source code would point to the wrong header files. ROS 2 convention :shrug:

vknisley-swri: OK, thanks for the clarification! I'll revert this and push the commit up.

DangitBen: I might be right but for the wrong reasons? Or maybe both are true? There's this page that explains some rationale:
https://colcon.readthedocs.io/en/released/user/overriding-packages.html#install-headers-to-a-unique-include-directory

PR #854, opened 2025-07-16
Fixing misnamed library in CMakeLists

danthony06:


danthony06:
@DangitBen does this look right?

PR #863, opened 2025-12-15
HiDPI Resolution Issues

laelliott:
I tend to run on 200% resolution since the way 4k renders on my monitors makes the text too small
Screenshot from 2025-12-15 13-43-13

This results in Mapviz only rendering in the bottom corner
Screenshot from 2025-12-15 13-40-32

It's apparently a known issue when people run monitors with a >100% scaling factor that the QT render only renders in the bottom left corner  

https://qt-project.atlassian.net/browse/QTBUG-59956
https://doc.qt.io/archives/qt-5.15/highdpi.html

With the update line it now renders properly. I tested in both 100 and 200% resolution
Screenshot from 2025-12-15 13-39-03

### DIFF
--- a/mapviz/src/map_canvas.cpp
+++ b/mapviz/src/map_canvas.cpp
@@ -592,7 +592,7 @@ void MapCanvas::UpdateView()
   if (initialized_) {
     Recenter();
 
-    glViewport(0, 0, width(), height());
+    glViewport(0, 0, width() * devicePixelRatio(), height() * devicePixelRatio());
     glMatrixMode(GL_PROJECTION);
     glLoadIdentity();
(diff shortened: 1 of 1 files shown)


laelliott:
Additionally, I'm working on a port from `QGLWidget` to `QOpenGLWidget` to solve more of these compatibility issues here: https://github.com/laelliott/mapviz/tree/lelliott/port-to-widget

matt-attack:
> Additionally, I'm working on a port from `QGLWidget` to `QOpenGLWidget` to solve more of these compatibility issues here: https://github.com/laelliott/mapviz/tree/lelliott/port-to-widget

Guess the GL just wasn't open enough.

PR #865, opened 2026-01-07
Added image rotation option for image_plugin

agyoungs:
This adds the ability to rotate image (increments of 90 degrees) in the image_plugin directly inside mapviz.


danthony06:
Blooming this as part of 2.6.1 now

PR #866, opened 2026-03-16
Fixed YAML parsing of doubles for image_plugin

agyoungs:
When using percent as the unit, image_plugin was failing because the width and height were being read as integers. This fixes that and makes sure to round pixels to the nearest integer


danthony06 (approved):

danthony06:
Just bloomed this fix in v2.6.2.

PR #871, opened 2026-05-06
Refactor OpenGL Interface

danthony06:
With the upcoming switch of Rolling from Ubuntu 24.04 to 26.04, it created a lot of problems because we will have to switch from Qt5 to Qt6. This is a first attempt at refactoring to use all native Qt OpenGL functionality and remove the Glew and Glut dependencies to simplify the dependency chain. Since the Qt5->6 change is going to break several things, I figured we should take the chance to make this change.

As a byproduct, this should also allow the system to run on Arm platforms like the Pi and Jetson.


danthony06:
@agyoungs, @rjb0026, @nick-alton, if any of you have time to look at this and test it, I'd appreciate it.

agyoungs:
> @agyoungs, @rjb0026, @nick-alton, if any of you have time to look at this and test it, I'd appreciate it.

I can probably take a look at this in about a week if that works for you. Obviously there's no reason to wait on my review if everyone else is happy with it before then

danthony06:
Sounds good. This change has been on my mind for a while, with 26.04 breaking Qt5 backwards compatibility anyway, I figured we might as well go all in. I'd especially like to get this running on a Pi or Jetson, because that's a pretty common request.

danthony06:
And I'll make this a 3.0 release if go forward with it because of the API change.

danthony06:
@jgassaway you might be interested in this as well.

mapviz/include/mapviz/mapviz.hpp
```diff
@@ -215,6 +218,12 @@ public Q_SLOTS:
   MapCanvas* canvas_;
   std::map<QListWidgetItem*, MapvizPluginPtr> plugins_;
 
+  // Config dock pin/auto-hide
+  QToolButton* pin_button_;
+  QLabel* title_label_;
+  QWidget* collapsed_label_;
+  bool config_panel_pinned_;
```
DangitBen:
```suggestion
  QToolButton* pin_button_ = nullptr;
  QLabel* title_label_ = nullptr;
  QWidget* collapsed_label_ = nullptr;
  bool config_panel_pinned_ = false;
```

mapviz/include/mapviz/rqt_mapviz.hpp
```diff
@@ -47,6 +59,16 @@
...
+#ifdef MAPVIZ_RESTORE_SLOTS_MACRO
+#pragma pop_macro("slots")
+#undef MAPVIZ_RESTORE_SLOTS_MACRO
+#endif
+
+#ifdef MAPVIZ_RESTORE_SIGNALS_MACRO
+#pragma pop_macro("signals")
+#undef MAPVIZ_RESTORE_SIGNALS_MACRO
+#endif
+
```
DangitBen, line 71: I see the comment above, but is all this still necessary? ROS generally doesn't use boost anymore in more core stuff.

mapviz/launch/mapviz.launch.py
```diff
@@ -5,12 +5,15 @@
...
+        on_exit=launch.actions.Shutdown(),
+    )
+
     return launch.LaunchDescription([
-        launch_ros.actions.Node(
-            package="mapviz",
-            executable="mapviz",
-            name="mapviz",
-        ),
+        mapviz_node,
```
DangitBen, line 16: Why do this for only the `mapviz_node`?

mapviz/src/map_canvas.cpp
```diff
@@ -47,7 +47,20 @@
...
+{
+QSurfaceFormat CreateMapCanvasFormat(bool enable_antialiasing)
+{
+  QSurfaceFormat format;
+  format.setRenderableType(QSurfaceFormat::OpenGL);
+  format.setProfile(QSurfaceFormat::CompatibilityProfile);
+  format.setVersion(2, 1);
+  format.setSwapBehavior(QSurfaceFormat::DoubleBuffer);
+  format.setSwapInterval(1);
+  format.setSamples(enable_antialiasing ? 4 : 0);
```
DangitBen, line 61: What is this magic number?

mapviz/src/map_canvas.cpp
```diff
@@ -47,7 +47,20 @@
...
 {
-
+namespace
+{
+QSurfaceFormat CreateMapCanvasFormat(bool enable_antialiasing)
+{
+  QSurfaceFormat format;
+  format.setRenderableType(QSurfaceFormat::OpenGL);
+  format.setProfile(QSurfaceFormat::CompatibilityProfile);
+  format.setVersion(2, 1);
```
DangitBen, line 58: What are these magic numbers?

mapviz/src/map_canvas.cpp
```diff
@@ -146,15 +162,15 @@ void MapCanvas::InitializePixelBuffers()
...
-      glBufferDataARB(GL_PIXEL_PACK_BUFFER_ARB, buffer_size, 0, GL_STREAM_READ_ARB);
-      glBindBufferARB(GL_PIXEL_PACK_BUFFER_ARB, pixel_buffer_ids_[1]);
-      glBufferDataARB(GL_PIXEL_PACK_BUFFER_ARB, buffer_size, 0, GL_STREAM_READ_ARB);
-      glBindBufferARB(GL_PIXEL_PACK_BUFFER_ARB, 0);
+      glGenBuffers(2, pixel_buffer_ids_);
+      glBindBuffer(GL_PIXEL_PACK_BUFFER, pixel_buffer_ids_[0]);
+      glBufferData(GL_PIXEL_PACK_BUFFER, buffer_size, 0, GL_STREAM_READ);
+      glBindBuffer(GL_PIXEL_PACK_BUFFER, pixel_buffer_ids_[1]);
+      glBufferData(GL_PIXEL_PACK_BUFFER, buffer_size, 0, GL_STREAM_READ);
+      glBindBuffer(GL_PIXEL_PACK_BUFFER, 0);
```
DangitBen, line 174: Here's where I'd recommend using `std::array`, so that you can use the `size` method instead of hard-coding.

As an alternative I'd probably recommend either having a named constant in the header for the length.
Or adding something like this:
```c++
size_t pixel_buffer_length = sizeof(pixel_buffer_ids_) / sizeof(pixel_buffer_ids_[0]);
glGenBuffers(pixel_buffer_length, pixel_buffer_ids_);
```

mapviz/src/map_canvas.cpp
```diff
@@ -216,19 +232,19 @@ void MapCanvas::CaptureFrame(bool force)
...
     int32_t next_index = (pixel_buffer_index_ + 1) % 2;
 
-    glBindBufferARB(GL_PIXEL_PACK_BUFFER_ARB, pixel_buffer_ids_[pixel_buffer_index_]);
+    glBindBuffer(GL_PIXEL_PACK_BUFFER, pixel_buffer_ids_[pixel_buffer_index_]);
     glReadPixels(0, 0, width(), height(), GL_BGRA, GL_UNSIGNED_BYTE, 0);
-    glBindBufferARB(GL_PIXEL_PACK_BUFFER_ARB, pixel_buffer_ids_[next_index]);
+    glBindBuffer(GL_PIXEL_PACK_BUFFER, pixel_buffer_ids_[next_index]);
     GLubyte* data = reinterpret_cast<GLubyte*>(
-      glMapBufferARB(GL_PIXEL_PACK_BUFFER_ARB, GL_READ_ONLY_ARB));
+      glMapBuffer(GL_PIXEL_PACK_BUFFER, GL_READ_ONLY));
```
DangitBen: I think you should avoid `reinterpret_cast` when possible. `void*` can be `static_cast`'ed instead.

Either way I'd recommend one of the following (`auto` to avoid type redundancy):
```c++
    auto data = reinterpret_cast<GLubyte*>(
      glMapBuffer(GL_PIXEL_PACK_BUFFER, GL_READ_ONLY));
```

(Preferred)
```c++
    auto data = static_cast<GLubyte*>(
      glMapBuffer(GL_PIXEL_PACK_BUFFER, GL_READ_ONLY));
```

mapviz/src/mapviz.cpp
```diff
@@ -82,6 +86,35 @@
...
+class VerticalLabel : public QWidget
+{
+public:
+  explicit VerticalLabel(const QString& text, QWidget* parent = nullptr)
+    : QWidget(parent), text_(text) {
+    setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Expanding);
+  }
+  QSize sizeHint() const override {
+    QFontMetrics fm(font());
+    return QSize(fm.height() + 4, fm.horizontalAdvance(text_) + 8);
```
DangitBen: What are these magic numbers?

mapviz/src/mapviz.cpp
```diff
@@ -122,6 +159,44 @@ Mapviz::Mapviz(bool is_standalone, int argc, char** argv, QWidget *parent, Qt::W
...
+    pin_button_->setCheckable(true);
+    pin_button_->setChecked(true);
+    pin_button_->setToolTip("Pin panel (unpin to auto-hide)");
+    pin_button_->setIcon(QIcon::fromTheme("window-pin",
+      QIcon::fromTheme("object-locked")));
+    pin_button_->setAutoRaise(true);
+    pin_button_->setFixedSize(20, 20);
+    // Use text fallback if no icon theme available
+    if (pin_button_->icon().isNull()) {
+      pin_button_->setText("\xF0\x9F\x93\x8C");  // pin emoji as fallback
```
DangitBen, line 196: 1. You using an emoji here is funny
2. Why use the code for an emoji here over just actually putting the emoji? 📌


```suggestion
      pin_button_->setText("📌");  // pin emoji as fallback
```

mapviz/src/mapviz.cpp
```diff
@@ -1288,6 +1363,33 @@ void Mapviz::ToggleConfigPanel(bool on)
...
 
+void Mapviz::TogglePinConfigPanel(bool pinned)
+{
+  config_panel_pinned_ = pinned;
+  if (pinned) {
+    pin_button_->setToolTip("Panel pinned (click to auto-hide)");
+    // Restore full dock
+    title_label_->setText("Config");
+    ui_.configdock->setMaximumWidth(QWIDGETSIZE_MAX);
+    ui_.configdock->setMinimumWidth(332);
```
DangitBen: How was this magic number chosen?

mapviz/src/mapviz.cpp
```diff
@@ -1553,4 +1655,30 @@ void Mapviz::HandleProfileTimer()
...
 }
+
+bool Mapviz::eventFilter(QObject* object, QEvent* event)
+{
+  if (object == ui_.configdock && !config_panel_pinned_) {
+    if (event->type() == QEvent::Enter) {
+      // Expand on mouse enter
+      title_label_->setText("Config");
+      ui_.configdock->setMaximumWidth(QWIDGETSIZE_MAX);
+      ui_.configdock->setMinimumWidth(332);
```
DangitBen: Same question about magic number

mapviz/src/mapviz.cpp
```diff
@@ -1553,4 +1655,30 @@ void Mapviz::HandleProfileTimer()
...
+      ui_.widget->show();
+    } else if (event->type() == QEvent::Leave) {
+      // Collapse on mouse leave
+      title_label_->setText("");
+      ui_.widget_2->hide();
+      ui_.configs->hide();
+      ui_.widget->hide();
+      collapsed_label_->setVisible(true);
+      ui_.configdock->setMinimumWidth(28);
+      ui_.configdock->setMaximumWidth(28);
```
DangitBen: More magic numbers. If these are pixel values, usually you'd prefer percentages or making it based on something tangible like font/element size.

mapviz/CMakeLists.txt
```diff
@@ -1,41 +1,57 @@
...
 if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
-### END CMAKE POLICIES ###
 
 project(mapviz 
   DESCRIPTION "Mapviz is a ROS based visualization tool with a plug-in system similar to RVIZ focused on visualizing 2D data."
   LANGUAGES CXX)
 
+function(mapviz_ensure_interface_alias alias_target dependency_target)
```
DangitBen, line 17: Some usage documentation on these functions might make things more readable. Also maybe moving the functions out into a separate `.cmake` file that gets imported to keep the project configuration code more concise.

mapviz/CMakeLists.txt
```diff
@@ -1,41 +1,57 @@
...
+  if(TARGET ${alias_target})
+    return()
+  endif()
+
+  add_library(${alias_target} INTERFACE IMPORTED)
+  set_target_properties(${alias_target} PROPERTIES
+    INTERFACE_LINK_LIBRARIES ${dependency_target})
+endfunction()
+
+function(mapviz_ensure_opencv_component component)
```
DangitBen, line 29: Why do we have a duplicate of this function? `mapviz-extras.cmake` already has a version of this function.

mapviz/CMakeLists.txt
```diff
@@ -45,133 +61,199 @@ find_package(tf2 REQUIRED)
...
+if(Qt5_FOUND)
+  set(MAPVIZ_QT_VERSION_MAJOR 5)
+else()
+  find_package(Qt6 REQUIRED COMPONENTS Concurrent Core Gui OpenGL Widgets)
+  set(MAPVIZ_QT_VERSION_MAJOR 6)
+endif()
+
+if(MAPVIZ_QT_VERSION_MAJOR EQUAL 6)
+  find_package(Qt6 REQUIRED COMPONENTS OpenGLWidgets)
+endif()
```
DangitBen: Why isn't this combined with the `else` branch that precedes it??

mapviz/CMakeLists.txt
```diff
@@ -45,133 +61,199 @@ find_package(tf2 REQUIRED)
...
-### Qt5 on ROS2 ###
-qt5_add_resources(RCC_SRCS src/resources/icons.qrc)
-qt5_wrap_ui(UI_SRC_FILES ${UI_FILES})
-qt5_wrap_cpp(MOC_FILES ${HEADER_FILES})
+set(MAPVIZ_QT_LIBRARIES
+  Qt${MAPVIZ_QT_VERSION_MAJOR}::Concurrent
+  Qt${MAPVIZ_QT_VERSION_MAJOR}::Core
+  Qt${MAPVIZ_QT_VERSION_MAJOR}::Gui
+  Qt${MAPVIZ_QT_VERSION_MAJOR}::OpenGL
+  Qt${MAPVIZ_QT_VERSION_MAJOR}::Widgets
```
DangitBen, line 141: Feels like this could be done better with aliases:
Earlier when you set  `MAPVIZ_QT_VERSION_MAJOR` you could instead:
```cmake
add_library(mapviz::qt ALIAS Qt5)`
```

Then in these lines you could do (I think... I'm not certain if referencing a component off of that works the way I think):
```cmake
set(MAPVIZ_QT_LIBRARIES
     mapviz::qt::Concurrent
     ...
)
```

mapviz/CMakeLists.txt
```diff
@@ -45,133 +61,199 @@ find_package(tf2 REQUIRED)
...
+    Qt5Gui
+    Qt5OpenGL
+    Qt5Widgets)
+endif()
 
-### Build mapviz as an rqt plugin ###
-add_library(rqt_${PROJECT_NAME}
-  ${MOC_FILES}
+### Build mapviz core library ###
+add_library(${PROJECT_NAME}_core SHARED
```
DangitBen, line 163: Why rename this? Are you sure this doesn't break downstream packages that might rely on this name?

mapviz_plugins/cmake/mapviz_plugins-extras.cmake
DangitBen, line 1: Seems like another duplicated set of functions. Can't this be imported from mapviz? Especially since it exports them?

mapviz_plugins/src/attitude_indicator_plugin.cpp
```diff
@@ -334,7 +394,7 @@ namespace mapviz_plugins
     for (int i = 1; i < divisions; i++)
     {
       glVertex2d(-0.2 * std::cos(M_PI * i / divisions),
-                 -0.2 * std::sin(M_PI * i / divisions));
+                     -0.2 * std::sin(M_PI * i / divisions));
```
DangitBen: This seems unintentional.

multires_image/src/tile.cpp
```diff
@@ -188,18 +187,25 @@ namespace multires_image
   {
     if (!m_failed)
     {
-      if (m_textureLoaded)
+      if (m_textureLoaded && m_texture)
       {
-        glBindTexture(GL_TEXTURE_2D, m_textureId);
+        initializeOpenGLFunctions();
```
DangitBen: This is called in a lot of places, including most/all of the plugins. It only needs to be called once per ` QOpenGLFunctions`

multires_image/CMakeLists.txt
```diff
@@ -1,54 +1,90 @@
...
+if(Qt5_FOUND)
+  set(MULTIRES_IMAGE_QT_VERSION_MAJOR 5)
+else()
+  find_package(Qt6 REQUIRED COMPONENTS Core Gui OpenGL Widgets)
+  set(MULTIRES_IMAGE_QT_VERSION_MAJOR 6)
+endif()
+
+if(MULTIRES_IMAGE_QT_VERSION_MAJOR EQUAL 6)
+  find_package(Qt6 REQUIRED COMPONENTS OpenGLWidgets)
+endif()
```
DangitBen: Why is this not combined with the else above?

multires_image/CMakeLists.txt
```diff
@@ -59,67 +95,122 @@ target_include_directories(${PROJECT_NAME}
...
+  src/QGLMap.cpp
+)
+set(WIDGET_HEADER_FILES
+  include/${PROJECT_NAME}/QGLMap.hpp
+)
 
 add_library(multires_widget
   ${WIDGET_SRC_FILES}
+  ${WIDGET_HEADER_FILES}
+  src/QGLMap.ui
```
DangitBen, line 127: Why are these being added like source files to this library?

multires_image/CMakeLists.txt
```diff
@@ -59,67 +95,122 @@ target_include_directories(${PROJECT_NAME}
...
-add_executable(multires_view_node ${VIEW_NODE_SRC_FILES})
+set(VIEW_NODE_SRC_FILES
+  src/nodes/multires_view_node.cpp
+)
+set(VIEW_NODE_HEADER_FILES
+  include/${PROJECT_NAME}/multires_view_node.hpp
+)
+add_executable(multires_view_node
+  ${VIEW_NODE_SRC_FILES}
+  ${VIEW_NODE_HEADER_FILES})
```
DangitBen, line 157: Why is this being added like a source file instead of a `target_include_directories` call?

tile_map/include/tile_map/texture_cache.hpp
```diff
@@ -27,27 +27,31 @@
...
 {
   class Texture
   {
   public:
-    Texture(int32_t texture_id, size_t hash);
+    Texture(std::unique_ptr<QOpenGLTexture> texture, size_t hash);
     ~Texture();
 
-    const int32_t id;
+    QOpenGLTexture* GetTexture() const { return texture_.get(); }
```
DangitBen: This seems like a great way to lose most of the benefits of using a managed pointer. Why not return a reference to the unique pointer instead?

tile_map/CMakeLists.txt
```diff
@@ -30,30 +32,41 @@ find_package(swri_math_util REQUIRED)
...
+else()
+  find_package(Qt6 REQUIRED COMPONENTS Core Gui Network OpenGL Widgets)
+  set(TILE_MAP_QT_VERSION_MAJOR 6)
+endif()
 
-### PkgConfig ###
-find_package(PkgConfig REQUIRED)
+if(TILE_MAP_QT_VERSION_MAJOR EQUAL 6)
+  find_package(Qt6 REQUIRED COMPONENTS OpenGLWidgets)
+endif()
```
DangitBen: Why not add this to the else branch above?

tile_map/CMakeLists.txt
```diff
@@ -30,30 +32,41 @@ find_package(swri_math_util REQUIRED)
...
 
-### Use PkgConfig to find jsoncpp ###
-pkg_check_modules(JSONCPP REQUIRED jsoncpp)
+find_package(yaml-cpp QUIET CONFIG)
+if(NOT TARGET yaml-cpp)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(YamlCpp REQUIRED IMPORTED_TARGET yaml-cpp)
+  tile_map_ensure_interface_alias(yaml-cpp PkgConfig::YamlCpp)
+endif()
```
DangitBen, line 60: Aren't the both already required/guaranteed by mapviz?

DangitBen (changes requested):
Some small changes given the size of the PR. I think there's a lot of CMake here that I can't really validate at a glance. I think a lot of that is going to depend on it building and running on all combinations of OSes, platforms, compilers, and install//source setups we want to support.

mapviz/include/mapviz/rqt_mapviz.hpp
danthony06: I think we could still have someone downstream linking boost. It seems unlikely to be needed, but I also don't think it's hurting anything.

mapviz/launch/mapviz.launch.py
danthony06: I was having issues with it not shutting down everything down the chain. This propagates the shutdown through the chain.

mapviz/src/map_canvas.cpp
danthony06: Added comments

tile_map/CMakeLists.txt
danthony06: Simplified logic.

tile_map/CMakeLists.txt
danthony06: It removes the need to require Mapviz to transitively supply these. Mapviz could remove those requirements, but we still require them.

mapviz/src/mapviz.cpp
danthony06: Sometimes I don't have fancy editors that can render emojis.

mapviz/CMakeLists.txt
danthony06: I agree, but I think it's enough of a style thing we'll defer it.

mapviz/CMakeLists.txt
danthony06: This was to resolve namespace collisions that occurred while loading plugins with RQT.

mapviz/CMakeLists.txt
danthony06: In theory so we can build them independently

mapviz_plugins/cmake/mapviz_plugins-extras.cmake
danthony06: This lets us build these packages independently in the future.

multires_image/src/tile.cpp
danthony06: Okay, I simplified the init logic. Good catch.

multires_image/src/tile.cpp
```diff
@@ -140,6 +140,8 @@ namespace multires_image
     {
       m_mutex.lock();
 
+      initializeOpenGLFunctions();
```
DangitBen, line 143: This is happening on every texture load still?

danthony06: It's guarded by m_textureLoaded check in the conditional. It should only be called until a texture is successfully loaded.

danthony06:
@DangitBen I think I've addressed all your comments.

danthony06:
@nick-alton, @rjb0026, @agyoungs: @DangitBen and I are thinking about merging this and blooming it only to `rolling` because of the significant changes. This will also hopefully make the buildfarm happy so that we aren't blocked on there any more.

danthony06:
Okay, I'm releasing this to `rolling` as `3.0.0`. All other releases will stay on `2.6.4` until we get feedback on this. I do want to see if it makes the repo build on the buildfarm.

PR #876, opened 2026-05-28
Fixing bad cherry pick

danthony06:


DangitBen:
I'm not sure about these changes. Why are those additions causing issues?

danthony06:
This is backporting the RQT fix to the kilted-devel branch until we get 3.0 released to all ROS distros. I made a bad cherry pick and accidentally pushed directly to the main repo, so this is fixing that cherry pick.

DangitBen (approved):
I can't test this ~~but after looking through the cmakelists.txt it _seems_ fine.~~ The build issue is still a problem.

DangitBen (changes requested):
Need to fix build issue with bad include.

https://github.com/danthony06/mapviz/blob/rqt_fix/tile_map/include/tile_map/stadia_source.h#L34

Needs to change from:

```cpp
#include <tile_map/tile_source.hpp>
```

To:
```cpp
#include "tile_source.h"
```

danthony06:
@DangitBen should be good to go now

DangitBen (approved):
Looks good!

PR #877, opened 2026-06-17
Tilemap Plugin Peformance/Stability Improvements

agyoungs:
I've been having more and more issues with mapviz crashing as I've been using it for more simulation intensive (especially GPU) based projects. Most of my issues went away when disabling the tile_map plugin so this PR is an attempt to address the problems I was running into. The main issues I'm addressing (both occurring only when tile_map plugin was enabled):

-  Mapviz crashes
- Temporarily unresponsive Mapviz window

Summary of Changes:
- **Fixed seg fault causing mapviz to crash**: The most important fix was moving `std::sort(images.begin(), images.end(), ComparePriority);` inside the mutex check. Without that mapviz would crash on me pretty consistently when under heavy load. This is because there was a comparison check in another thread that was accessing out of bounds memory if checking the unprocessed queue while it was being sorted.
- Lowered the priority of the cache thread. I'm not sure how much performance improvement I got from doing this, but it at least should prioritize the main thread and prevent it from getting starved and freezing the mapviz window. If you're having performance issues, you would probably rather the image cache thread get starved instead of the main thread.
- Moved image decode entirely off the main thread into CacheThread. Similar to above, it seemed better to handle decoding in the cache thread so the main thread isn't tied up doing this
- Added step to remove unprocessed requests that aren't visible in the current frame (or at least requested by pre-cache). This shouldn't affect pre-cache as it gets requested with each SetView change. But it does reduce the backlog of requests that queue up as a user pans around. With aggressive panning and zooming, a user can create a queue of 5,000 + requests. In the old code the priority naturally works out such that the currently visible tiles get priority, but this still seemed wasteful to me. Instead I set it to remove any unprocessed tiles from the queue that aren't a part of the current SetView frame. This significantly reduces how much backlog can accumulate. The sideffect is that image tiles seem to load more quickly (slightly), but at the cost that the intermediate tiles when panning around don't get queued (i.e. if a user scrolls out a lot, only the previous view and the current zoomed out view will likely get fetched; all tiles in-between likely won't be fetched until you zoom back in to the intermediate layer). If this is not desirable behavior I can remove it.

### DIFF
--- a/tile_map/include/tile_map/image_cache.hpp
+++ b/tile_map/include/tile_map/image_cache.hpp
@@ -53,7 +53,7 @@ namespace tile_map
   class Image
   {
   public:
-    Image(const QString& uri, size_t uri_hash);
+    Image(const QString& uri, size_t uri_hash, uint64_t priority = 0);
     ~Image() = default;
 
     QString Uri() const { return uri_; }
@@ -67,14 +67,11 @@ namespace tile_map
     void AddFailure();
     bool Failed() const { return failed_; }
 
-    void SetBasePriority(int32_t priority) { base_priority_ = priority; }
-    int32_t BasePriority() const { return base_priority_; }
-
     void SetLastRequestedFrame(uint64_t frame) { last_requested_frame_ = frame; }
     uint64_t LastRequestedFrame() const { return last_requested_frame_; }
 
-    void SetLastRequestedTick(uint64_t tick) { last_requested_tick_ = tick; }
-    uint64_t LastRequestedTick() const { return last_requested_tick_; }
+    void SetPriority(uint64_t priority) { priority_ = priority; }
+    uint64_t Priority() const { return priority_; }
 
     bool Loading() const { return loading_; }
     void SetLoading(bool loading) { loading_ = loading; }
@@ -91,9 +88,8 @@ namespace tile_map
     bool loading_;
     int32_t failures_;
     bool failed_;
-    int32_t base_priority_ = 0;
     uint64_t last_requested_frame_ = 0;
-    uint64_t last_requested_tick_ = 0;
+    uint64_t priority_ = 0;
 
     mutable std::shared_ptr<QImage> image_;
     QByteArray pending_data_;
--- a/tile_map/include/tile_map/image_cache.hpp
+++ b/tile_map/include/tile_map/image_cache.hpp
@@ -31,10 +31,10 @@
 #define TILE_MAP_IMAGE_CACHE_HPP_
 
 #include <string>
-#include <limits>
 
 #include <rclcpp/logger.hpp>
 
+#include <QByteArray>
 #include <QCache>
 #include <QImage>
 #include <QMap>
@@ -53,7 +53,7 @@ namespace tile_map
   class Image
   {
   public:
-    Image(const QString& uri, size_t uri_hash, uint64_t priority = 0);
+    Image(const QString& uri, size_t uri_hash);
     ~Image() = default;
 
     QString Uri() const { return uri_; }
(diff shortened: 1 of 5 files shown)


DangitBen:
Thanks for the contribution @agyoungs!

This looks like a very useful PR! I'd like to test them out though (hopefully soon), because that is a significant set of changes to the plugin.

I saw that same behavior where the `tile_map` would crash when under heavy load, especially with detailed maps and fast panning/zooming. Great catch on the sorting being outside the lock.

As for the intermediate tiles not being queued, I think that is fine. That seems in line with other mapping tools, though they may hide it better with requesting an area slightly larger than the current view or using a basemap.

laelliott:
I appreciate you reworking these issues! I was likely the original regression 😭

agyoungs:
@DangitBen Yeah no problem! And I'm still testing a few things myself and I'll mark it ready once I'm done tweaking this PR. I'm also happy to split this into multiple PRs if that makes it easier!

@laelliott I feel like this one might be an old issue, but maybe it's more noticeable/exposed with the new Carto tile source?

agyoungs:
@DangitBen @danthony06 Also, do you plan on blooming 3.0.1+ for jazzy? I've had a chance to test from the `ros2-devel` branch on jazzy with the new OpenGL updates and it looks like everything's working well. I just had to change a few includes (.h to .hpp) in some custom mapviz plugins, but it works great!

Or should I add a patch for a jazzy branch separately for this issue?

danthony06:
@agyoungs yeah, if the testing has gone well for you then I think we can bloom this for all releases. We are going to need to fork for Lyrical and newer releases because prior releases built RQT against Qt5, and Lyrical and newer are built against Qt6, which means we'll need to have separate Qt5 and Qt6 versions of Mapviz so that the application can link against the appropriate version of Qt.

As far as releasing, I think it makes sense to merge this fix into both the `kilted-devel` and `ros2-devel` branches because it fixes such a nasty bug. I can then make a release with just the bug fix in it so that people can have that. I can then make the big Qt6 and OpenGL change so that there's a clean break there.

danthony06:
I'm also working on getting all dependencies released so we can build on 26.04 (Lyrical and Rolling). There were some breaking changes that I need to fix for `marti_common`, and then it should be working. I think the other dependencies are currently gettable via `rosdep`, and `marti_common` is the remaining blocker.

agyoungs:
> As far as releasing, I think it makes sense to merge this fix into both the kilted-devel and ros2-devel branches because it fixes such a nasty bug. I can then make a release with just the bug fix in it so that people can have that. I can then make the big Qt6 and OpenGL change so that there's a clean break there.

The first time I read this I don't think I fully comprehended that there was a `kilted-devel` branch. Is `ros2-devel` the rolling branch and then for the various ROS distros you're handling with a separate branch? Or do you anticipate getting rid of the `kilted-devel` branch now that there's some feedback on the OpenGL updates in jazzy (tbf it's not kilted so maybe not valid). I see that you're planning on maintaining two branches for qt5/6 transition which makes sense, but I just wanted to know if you need me to open up additional PRs or is this the only one that's needed?

danthony06:
Yeah, that branches are a bit of a mess right now. I had originally planned on blooming all distros out of `ros2-devel`, and v3.x would indicate our OpenGL change. Then when `lyrical` released, I found out that Qt6 is going to be a permanent breaking change that I'll need separate branches for. My tentative plan is:

1. Going forward, `humble`, `jazzy`, and `kilted` will all get released from the `kilted-devel` branch.
2. `lyrical` and future versions will release from the `ros2-devel` branch, at least until another breaking change happens.
3. It sounds like the TSC is thinking about doing something different with `rolling` going forward, but I don't know all the details. If possible, I'm going to keep blooming it from `ros2-devel` as well.
4. The Qt6 break threw a wrench in my version numbers. When I release the OpenGL changes, I'd like to still bump the major version number since the .hpp rename breaks compatibility. Right now, I'm thinking of blooming up through `kilted` as a 3.x release, and `lyrical` and newer go to 4.x

For where to merge, I'd say choose whether `ros2-devel` or `kilted-devel` is easier. I'm happy to port the change to the other branch if you want, and I won't say no if you want to make the change in both branches.

DangitBen:
I finally got a moment to test this and it works well. Definitely much more stable, especially when zooming/panning quickly. There are still _some_ moments where the view hitches, but I can't tell if that's better or worse than the old performance. I'll still happily advocate to merge for the stability though.

I didn't notice intermediate files not being queued. I could see new tiles being filled in while I was panning rapidly (while not releasing the mouse button).

agyoungs:
> I finally got a moment to test this and it works well. Definitely much more stable, especially when zooming/panning quickly. There are still _some_ moments where the view hitches, but I can't tell if that's better or worse than the old performance. I'll still happily advocate to merge for the stability though.
> 
> I didn't notice intermediate files not being queued. I could see new tiles being filled in while I was panning rapidly (while not releasing the mouse button).

Sorry, just getting back to this...

Yeah, I've had a few freezes myself since. It definitely fixed the crashing issue though. I suspect this plugin could be due for a bigger refactor to make it less of a performance hog

PR #878, opened 2026-06-19
Fixed bad initial path import for multires image plugin

agyoungs:
If you add the multires image plugin and save your config (or just allow mapviz to restore the last config) it fails to load because the default path is an empty string. This fixes that to basically just ignore the empty path.


DangitBen:
This also looks good to me.

PR #880, opened 2026-06-19
Plugin panel eliding

agyoungs:
Elides the plugin title text so that the enable checkboxes are always visible and easier to use on a narrow panel.

| Before | After |
| ---------- | ------- |
| Screenshot_2026-06-19_02-11-43 | Screenshot_2026-06-19_02-10-53 |


DangitBen:
I also notice that this keeps other elements, like in the marker plugin, on the screen more often.

Looks good to me.

PR #881, opened 2026-06-19
Persistent config dock size

agyoungs:
Quick summary:
- Moved config panel toggle to the left. It seems easier to click here because when it expands the config panel the button doesn't move. So it's easier to toggle expanded/collapsed without having to move the mouse
- Made the config panel size a persistent value. So when you re-expand it expands to the previous set value. Also when you save a configuration it exports the panel width and uses it on import instead of using the default panel width size.

| Before | After |
| ---------- | ------- |
| Screenshot_2026-06-19_02-50-29 | Screenshot_2026-06-19_02-47-04 |

### DIFF
--- a/mapviz/src/mapviz.cpp
+++ b/mapviz/src/mapviz.cpp
@@ -151,7 +151,8 @@ Mapviz::Mapviz(bool is_standalone, int argc, char** argv, QWidget *parent, Qt::W
     pin_button_(nullptr),
     title_label_(nullptr),
     collapsed_label_(nullptr),
-    config_panel_pinned_(true)
+    config_panel_pinned_(true),
+    pinned_panel_width_(CONFIG_PANEL_PINNED_WIDTH)
 {
   // Multiple users could be using mapviz, so its name needs to be anonymous,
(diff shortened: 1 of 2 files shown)


mapviz/src/mapviz.cpp
```diff
@@ -886,6 +896,9 @@ void Mapviz::Save(const std::string& filename)
       << ui_.actionShow_Capture_Tools->isChecked();
   out << YAML::Key << "window_width" << YAML::Value << width();
   out << YAML::Key << "window_height" << YAML::Value << height();
+  if (config_panel_pinned_) {
+    out << YAML::Key << "panel_width" << YAML::Value << ui_.configdock->width();
+  }
```
DangitBen, line 901: If this check fails and the config is saved, (or using the default-saved `~/.mapviz_config` file) it will cause the pinned width to be forgotten.

I'm not fully opposed, but I guess I don't see why we should go out of our way to forget it?

agyoungs: Ah, this is a good catch. I don't remember why I put this in there but I was probably messing around with something. I'll submit a follow up PR to fix this

PR #882, opened 2026-07-01
Robot Model Plugin

agyoungs:
I've been sitting on this one a while, but finally got some motivation to clean it up and get it upstream.

This plugin simply visualizes a Robot Model URDF via a file or topic (i.e. /robot_description). It takes advantage of OpenGL's Framebuffer Objects to do the heavy lifting of rasterizing the image once when a new source is detected, and any time the zoom level necessitates increasing the resolution. All of the rasterization baking is done on a background lower priority thread to help reduce latency on the main thread. A max allowable size of 1.5x the larger canvas dimension prevents the Framebuffer object from taking up too much memory. Any time the model is off screen it reduces the object to a minimal size so that we're not chewing up additional memory unnecessarily. You can also set the alpha to make the model more transparent.

For simple robot models, it's probably excessive to handle all the baking in a background buffer. But, after dealing with some intricate mesh models with 800K+ faces, this seemed like an effective strategy to make rendering efficient.

Screenshot_2026-07-01_13-48-26

Note: It does add dependencies for assimp and URDF

### DIFF
--- a/mapviz_plugins/src/robot_model_plugin.cpp
+++ b/mapviz_plugins/src/robot_model_plugin.cpp
@@ -61,6 +61,7 @@
 #include <algorithm>
 #include <array>
 #include <cmath>
+#include <filesystem>
 #include <fstream>
 #include <functional>
 #include <limits>
@@ -585,6 +586,14 @@ void RobotModelPlugin::FileEdited() {
   file_path_ = path;
   if (file_path_.empty()) {
     PrintWarning("No file selected.");
+    return;
+  }
+  if (std::filesystem::path(file_path_).extension() == ".xacro") {
+    const std::string out =
+        std::filesystem::path(file_path_).replace_extension("").string();
+    PrintError("Cannot load .xacro files directly");
+    RCLCPP_ERROR(rclcpp::get_logger("robot_model_plugin"),
+                 "Run: xacro %s > %s", file_path_.c_str(), out.c_str());
     return;
   }
   std::ifstream ifs(file_path_);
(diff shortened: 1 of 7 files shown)


mapviz_plugins/src/robot_model_plugin.cpp
```diff
@@ -0,0 +1,1056 @@
...
+      return ament_index_cpp::get_package_share_directory(pkg) + "/" + rel;
+    } catch (...) {
+      return {};
+    }
+  }
+  if (uri.substr(0, 7) == "file://") {
+    return uri.substr(7);
+  }
+  return uri;
+}
```
danthony06, line 99: It might be worth trying to do this using std::filesystem. I know Lyrical changed ament_index_cpp to use std::filesystem and not raw strings, so we'll have to update it anyway.

agyoungs:
Just added a quick test here to show it working on various standard geometries. It may be worthwhile to exercise this on common meshes you use to make sure it handles everything as expected. One limitation using the Lambert reflectance algorithm is that faces with the same perspective (i.e. a square pyramid pointed directly up), will all get the same shading. In the case of a square pyramid, it'll look exactly like a flat square. 

Screenshot_2026-07-01_14-41-53



And here is the urdf.xacro and meshes I used for this simple test:

[test_urdf.tar.gz](https://github.com/user-attachments/files/29566686/test_urdf.tar.gz)


I did extensive testing with much more intricate meshes and URDFs as well

danthony06 (approved):
We'll probably need to do some work with the files for Lyrical, but that can wait.

PR #884, opened 2026-07-01
Improved tf handling

agyoungs:
I kept noticing that when I set the target_frame to base_link that the mapviz canvas stuttered a lot. Putting the whole mapviz node in its own spin thread seemed to fix all these issues I was having.

For this PR I used Claude to diagnose and fix the problem, and I haven't had a chance to scrutinize it. I'm going to go ahead and open it for others to look at if I don't get back to it first.

### DIFF
--- a/mapviz/include/mapviz/mapviz.hpp
+++ b/mapviz/include/mapviz/mapviz.hpp
@@ -55,6 +55,7 @@
 
 // ROS libraries
 #include <rclcpp/rclcpp.hpp>
+#include <rclcpp/executors/single_threaded_executor.hpp>
 #include <pluginlib/class_loader.hpp>
 #include <tf2_ros/buffer.h>
 #include <tf2_ros/transform_listener.h>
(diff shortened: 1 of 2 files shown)


agyoungs:
In draft right now, because every once in a while the TFs for other plugins are out of sync by one frame. There are several other thread safety concerns that need addressing as well. I'll take a look at this tomorrow, but I don't think this should be merged as it stands.

danthony06:
Yeah, seems like a decent approach, but something we'd need to dig into deeper. @rjb0026 did you ever try something like this when you were looking at Mapviz performance problems?

danthony06:
I wonder if we need to do `spin_some()`, like RViz does: https://github.com/ros2/rviz/pull/1567/changes.

rjb0026:
> Yeah, seems like a decent approach, but something we'd need to dig into deeper. @rjb0026 did you ever try something like this when you were looking at Mapviz performance problems?

No, I really only messed with the TF2 API calls themselves and none of it helped much.

agyoungs:
Closing as #887 will handle this

PR #885, opened 2026-07-10
Update for Deprecated Functionality

danthony06:
Fixes various ROS 2 and Qt warnings related to deprecated functionality in ROS 2 Lyrical and Qt 6.

### DIFF
--- a/mapviz/CMakeLists.txt
+++ b/mapviz/CMakeLists.txt
@@ -74,6 +74,17 @@ endif()
 
 find_package(geometry_msgs REQUIRED)
 find_package(image_transport REQUIRED)
+
+# image_transport (part of image_common) changed the ImageTransport
+# constructor from taking an rclcpp::Node::SharedPtr to taking an
+# rclcpp::Node& starting with image_common 6.4.0. Detect which API is
+# available so mapviz.cpp can call the correct one.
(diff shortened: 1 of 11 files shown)


danthony06:
@agyoungs I think this supersedes your PR about improved TF handling, and eliminates a bunch of warnings with Lyrical while keeping compatibility with older ROS distros.

agyoungs:
> @agyoungs I think this supersedes your PR about improved TF handling, and eliminates a bunch of warnings with Lyrical while keeping compatibility with older ROS distros.

This looks good, but doesn't seem to fix the lagging tf issues I see in mapviz

danthony06:
Good point. Thinking about the `tf` lag, I'm starting to suspect that it's due to processing the TF message queue blocking the system from running, because it's all running on one thread. I suspect we need to properly multithread the system so the rendering runs on a separate thread from processing messages (like your work did), but it might benefit everything if we used a multithreaded executor so that we could process more data in parallel. We'll need to ensure everything is made properly thread safe though, which might take a bit of work.

danthony06:
I'll merge this for now, because I think no matter what we do to resolve the `tf` lag, we'll want these changes and it might make them simpler to implement.

PR #887, opened 2026-07-12
Separate ROS and UI Threads

danthony06:
This separates out the processing done on callbacks and the ROS interface to its own thread, and lets the UI run independently on a separate thread. This should hopefully make the application more responsive.

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -10,5 +10,6 @@ has_children: true
 
 | Title | Description |
 |:- |:- |
+| [Writing a plugin]({{site.basurl}}/mapviz/guides/writing_a_plugin) | How to create a mapviz display plugin for your own data types in your own package, including the threading contract plugins must follow. |
 | [Local tile_map imagery]({{site.basurl}}/mapviz/guides/local_tile_map_imagery) | How to wrangle data to the correct format and direct mapviz's `tile_map` plugin to use it from local storage. |
 | [Geospatial Imagery From USGS Earth Explorer]({{site.basurl}}/mapviz/guides/imagery_usgs_earth_explorer) | An example of how to get imagery that could be used as a source for local `tile_map` imagery. Provided for reference only. |
--- a/docs/guides/writing_a_plugin.md
+++ b/docs/guides/writing_a_plugin.md
@@ -1,0 +1,183 @@
+---
+title: Writing a plugin
+parent: Guides
+has_children: false
+---
+
+# Writing a plugin
+{: .no_toc }
+
+Mapviz displays are plugins loaded at runtime through
+[pluginlib](https://github.com/ros/pluginlib). The core application knows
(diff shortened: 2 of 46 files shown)


danthony06:
@agyoungs and @DangitBen I think this is a thread safe implementation of https://github.com/swri-robotics/mapviz/pull/884. It uses signals and slots for thread safety, and runs the ROS callbacks and Qt on separate threads. The structure of the plugin code has to change a little, but it doesn't seem too bad.

agyoungs (commented):
@danthony06 Thanks for looking at this! I definitely agree that the best solution is to keep the rendering on its own dedicated thread that can't get bogged down by subscription threads. I tested this branch and it works great!

One thing I worry about is how this will affect plugin developers. It would be nice if we can enforce the threading paradigm with the compiler.

What if we made `node_` private within `Mapviz Plugin`? And then we add some templated base-class functions that handle all the plumbing to keep things thread safe.

For subscriptions, maybe something like:

```
template <typename MsgT>
void Subscribe(const std::string& topic, const rmw_qos_profile_t& qos,
                typename rclcpp::Subscription<MsgT>::SharedPtr& out_sub,
                std::function<void(typename MsgT::ConstSharedPtr)> on_gui_thread);

template <typename MsgT, typename DecodedT>
void Subscribe(const std::string& topic, const rmw_qos_profile_t& qos,
                typename rclcpp::Subscription<MsgT>::SharedPtr& out_sub,
                DecodedT (*decode)(const typename MsgT::ConstSharedPtr&),  // <-- function pointer
                std::function<void(std::shared_ptr<DecodedT>)> on_gui_thread);
```

And then for anything not funneled through the helpers (config-widget slots, Draw(), Transform(), GetTransform()), add Q_ASSERT(QThread::currentThread() == thread()) at entry.

I think stuff like logging, clock access, params?, are safe as-is. We can add helpers for them, too, though.

We'll definitely need helpers for create_subscription, create_client, create_service (if we want to support it), create_wall_timer. I think create_publisher is safe, but we might as well make a helper for it and just keep node_ totally private for consistency.

I'm happy to submit a  PR to your branch if you think it's worthwhile. Or it's fine if you'd rather handle it. Just let me know!

agyoungs:
Of course a developer could always add an rclcpp::Node inside the plugin directly. But we can update the README for best practices regarding this.

danthony06:
I like those suggestions. I can take a shot at it tonight. To keep the changes manageable, I'll merge this in, and then we can add the helper utilities and thread checking.

We could also add a simple plugin skeleton to the documentation to help developers.

I did worry about downstream developers, it's a big change, but I don't think we can avoid making significantly breaking changes and get this paradigm in. Steve Macenski suggested making a post on ROS Discourse about this, which I think is a good idea.

Issue #888, opened 2026-07-13
Add Helper Functions for Plugins


danthony06:
Per @agyoungs suggestion in https://github.com/swri-robotics/mapviz/pull/887, we should add helper functions to help enforce the boundary between the ROS and Qt threads for plugins, as well as add assertions to help enforce the thread boundaries.

danthony06:
Closed via https://github.com/swri-robotics/mapviz/pull/890

PR #890, opened 2026-07-14
Add Helper Functions for New Threading Model

danthony06:
This adds helper functions that abstract away the knowledge of the underlying shared datatypes which will make it harder for plugin developers to inadvertently create memory-unsafe code when writing plugins. Addresses https://github.com/swri-robotics/mapviz/issues/888.


danthony06:
@agyoungs and @DangitBen here's the change to implement helper functions to hide the underlying node for thread safety reasons.

mapviz_plugins/src/plan_route_plugin.cpp
```diff
@@ -255,7 +255,7 @@ namespace mapviz_plugins
     initializeOpenGLFunctions();
     canvas->doneCurrent();
 
-    retry_timer_ = node_->create_wall_timer(1000ms, [this](){Retry();});
+    retry_timer_ = NodeUnsafe()->create_wall_timer(1000ms, [this](){Retry();});
```
agyoungs: It looks like create_wall_timer fires on the background ROS spin thread. Its callback calls Retry which calls PlanRoute which touches UI elements:

```
bool start_from_vehicle = ui_.start_from_vehicle->isChecked();
std::string service = ui_.service->text().toStdString();
```

I think this will be a problem...Maybe use a QT Timer instead? There's not really a reason to use a ros Timer here, right? Or just add a createTimer function helper?

agyoungs: Probably makes sense to add ` MAPVIZ_ASSERT_GUI_THREAD()` to the PlanRoute/Retry callbacks as well

mapviz_plugins/src/attitude_indicator_plugin.cpp
```diff
@@ -363,6 +331,7 @@ namespace mapviz_plugins
 
   void AttitudeIndicatorPlugin::Draw(double x, double y, double scale)
   {
+    MAPVIZ_ASSERT_GUI_THREAD();
```
agyoungs: I see this added to a lot of plugins, but can we centralize it to remove the onus from the developer?

It looks like Draw/Paint Plugin should call the overridden plugin class functions?
```
void DrawPlugin(double x, double y, double scale)
{
  if (visible_ && initialized_) {
    MAPVIZ_ASSERT_GUI_THREAD();
    meas_transform_.start();
    Transform();
    meas_transform_.stop();
    meas_draw_.start();
    Draw(x, y, scale);
    meas_draw_.stop();
  }
}

void PaintPlugin(QPainter* painter, double x, double y, double scale)
{
  if (visible_ && initialized_) {
    MAPVIZ_ASSERT_GUI_THREAD();
    ...
  }
}
```

And then of course these functions are always called from the GUI thread, so they kind of don't server a purpose I guess. Maybe still good to have it. I guess we could make Draw, Paint and Transform protected to really make sure that only mapviz and the plugin can all these functions. And if that's the case it will always be called by the GUI thread unless the developer does something intentionally "wrong".

mapviz_plugins/src/disparity_plugin.cpp
```diff
@@ -167,21 +159,23 @@ namespace mapviz_plugins
...
+        });
+
+      RCLCPP_INFO(Logger(), "Subscribing to %s", topic_.c_str());
     }
   }
   void DisparityPlugin::SelectTopic()
   {
     auto [topic, qos] = SelectTopicDialog::selectTopic(
-      node_,
+      NodeUnsafe(),
```
agyoungs: I think it might be nice to make `NodeUnsafe` uncommon for developers to ever use. And this is a pretty common pattern (getting a list of topics). So maybe something like this in MapvizPlugin to abstract the use away?

```
protected:
  template <typename... Args>
  auto SelectTopic(Args&&... args)
  {
    return SelectTopicDialog::selectTopic(NodeUnsafe(), std::forward<Args>(args)...);
  }
```

Or just providing API calls for getting the topic types directly (which is actually useful for the marker_plugin and textured_marker_plugin, too)

```
protected:
  mapviz_plugins::TopicSource TopicSource() const
  {
    return {
      [this] { return node_->get_topic_names_and_types(); },
      [this] { return node_->get_service_names_and_types(); },
      Logger()
    };
  }
```

mapviz_plugins/src/robot_model_plugin.cpp
```diff
@@ -647,28 +639,26 @@ void RobotModelPlugin::TopicEdited() {
 
   description_sub_.reset();
 
-  if (topic_.empty() || !node_) {
+  if (topic_.empty() || !NodeUnsafe()) {
```
agyoungs: I think this one is actually leftover from a mistake I committed. But I don't think we need the node_ check here. So I think we can delete the NodeUnsafe check.

mapviz_plugins/src/attitude_indicator_plugin.cpp
agyoungs, line 416: I know this isn't strictly necessary, but if we end up pushing the ASSERT_GUI_THREAD check into the Drawing functions, I guess it belongs in the Load/Save config options. If we want it centralized, it could go in a protected LoadConfigPlugin and SaveConfigPlugin function within MapvizPlugin that then calls the protected LoadConfig/SaveConfig for each plugin.

agyoungs (commented):
Thanks for doing this! It works well on my end. I left some comments

danthony06:
@agyoungs I think I got everything that's worth getting.

agyoungs (commented):
Everything LGTM!

Issue #770, opened 2022-12-06
Unable to rotate markers in MapViz


ALCarroll24:
Hi,

I am trying to display a covariance ellipse within MapViz. Because I did not find any native functionality within MapViz for this, I am trying to draw a covariance ellipse using Markers.

This approach works fine to create a viewable ellipse. However, rotating the ellipse seems to do nothing and thus I can't fully display covariance.

I have tried viewing the same marker messages within RVIZ, where rotation and stretching the ellipse seems to work fine.

Are marker rotations not implemented in MapViz and if not are there are any other approaches I can take for viewing covariance.

Thanks

malban:
> Are marker rotations not implemented in MapViz and if not are there are any other approaches I can take for viewing covariance.

It looks like the orientation isn't taken into account when drawing the ellipse for a cylinder marker: https://github.com/swri-robotics/mapviz/blob/master/mapviz_plugins/src/marker_plugin.cpp#L566-L575, which seems like an oversight.

The [odometry](https://swri-robotics.github.io/mapviz/plugins/odometry/) plugin can draw the position covariance for an odometry message.

danthony06:
https://github.com/swri-robotics/mapviz/pull/897

PR #897, opened 2026-07-16
Using marker rotation when rendering

danthony06:


danthony06:
Resolves https://github.com/swri-robotics/mapviz/issues/770.

Issue #804, opened 2023-11-29
Update Min/Max values when Auto is triggered For Pointcloud plugin 


jgassaway:
In RVIZ, when coloring pointclouds by intensity, when you click "Autocompute Intensity" (the equivalent of mapviz's "Use Auto Max/Min"), the values for min/max intensity will automatically update to the max/min of the current pointcloud. These values can then be manually scaled by disabling that option and modifying the min or max values. 
In mapviz, these values are not updated, so you must manually check the scale of your data by experimentation, or prior knowledge of the max/min bounds. 

It would be helpful to update the max/min values displayed in the pointcloud2 plugin when "Use Auto Max/Min" is toggled.

danthony06:
Fixed via https://github.com/swri-robotics/mapviz/pull/901.

PR #901, opened 2026-07-24
Improving autocalculation of pointcloud colors

danthony06:


danthony06:
Fixes https://github.com/swri-robotics/mapviz/issues/804. 

- Added UpdateAutoRange(): computes the min/max over buffered scans for the active color transformer, writes them to min_value_/max_value_, mirrors them into the minValue/maxValue spin boxes (signals blocked to avoid a recursive recolor), and returns whether the range moved. Skips flat color and unpack_rgb, where no range is in play.
- Range is now recomputed fresh rather than accumulated forever, giving per-cloud semantics — identical to RVIZ at the default buffer_size_ of 1, and shrinking back as outliers age out with a larger buffer.
- Extracted ColorScan() and made recoloring conditional on the range having moved — a shifted range repaints the whole buffer, a stable one costs only the new scan.
- handleScan() now pushes the new cloud into the buffer before computing the range, so the computed range matches the rendered point cloud
- scans blanked by ResetTransformedPointClouds() regain colors on the next message instead of staying invisible.

